PR #31, opened 2016-04-26
Rubocop errors

Anbranin:


app/controllers/maneuvers_controller.rb
```diff
@@ -18,7 +18,8 @@ def next_participant
                                         participant: participant.number)
       end
     else redirect_to maneuvers_path,
-         notice: 'There are no more participants in the queue for this maneuver.'
+                     notice: 'There are no more participants in the queue for
```
dfaulken: Won't this put a line break in the middle of the notice?

Anbranin: No
screen shot 2016-04-27 at 4 43 19 pm

PR #32, opened 2016-05-02
Names uniqueness alert

### LINKED ISSUE #28: Handle same participant names
... in a better way than a 400 page.

bgregg:
No more 404 page when you try to create a participant with the same name as one that already exists. Closes #28.


bgregg:
screen shot 2016-05-02 at 10 27 19 am

app/controllers/participants_controller.rb
```diff
@@ -12,10 +12,19 @@ def assign_number
...
-    PrivatePub.publish_to '/scoreboard', participant
+    participant = Participant.new user_params
+    if participant.save
+      redirect_to participants_path,
+        notice: 'Participant has been created.'
+      PrivatePub.publish_to '/scoreboard', participant
+    elsif participant.errors.include?(:name)
+      redirect_to participants_path,
+        alert: 'Participant with that name already exists!'
+    # add more failures here if necessary
```
dfaulken: It'd be good to do it more dynamically than this.

[Here](https://github.com/umts/jobapps/blob/master/app/controllers/application_controller.rb#L63) is an [example](https://github.com/umts/jobapps/blob/master/app/views/layouts/application.haml#L22-L27).

bgregg:
screen shot 2016-05-02 at 1 56 14 pm

PR #33, opened 2016-05-23
Added Bus, Circle Check Quiz, and Quiz Score Tests

AkivaGreen:
Added bus, circle check quiz, and quiz score tests. All the tests that fail are due to a user having access they are not supposed to have.


app/models/quiz_score.rb
```diff
@@ -5,6 +5,8 @@ class QuizScore < ActiveRecord::Base
 
   validates :participant, :points_achieved, :total_points, presence: true
   validates :participant, uniqueness: true
+  #validates :points_achieved, numericality: { less_than_or_equal_to: :total_points,
```
dfaulken: Whoops, forgot to uncomment that.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 22.65
+    "covered_percent": 34.71
```
dfaulken: 👍

dfaulken:
As a general comment, your indentation is wrong - you're doing one tab as opposed to two spaces. Please update the files you've added to be indented by two spaces per level.

spec/features/bus_delete_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+describe 'deleting a bus - admin' do
```
dfaulken: I might suggest structuring this test differently to remove some of the duplication:

```ruby
describe 'deleting a bus'
  before(:each) do
    some_setup
  end
  context 'as admin' do
    it 'works'
  end
  # etc.
end
```

spec/features/bus_delete_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+describe 'deleting a bus - admin' do 
+	before :each do 
+		create :bus
+		when_current_user_is :admin
+		visit buses_url
+	end
+	context 'with admin privelege' do 
+		it 'delete a bus' do
```
dfaulken: "It delete a bus" isn't quite right

spec/features/bus_delete_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+describe 'deleting a bus - admin' do 
+	before :each do 
+		create :bus
+		when_current_user_is :admin
+		visit buses_url
+	end
+	context 'with admin privelege' do 
+		it 'delete a bus' do 
+			expect(page).to have_text 'Bus 1'
```
dfaulken: Better to bind the bus to some variable, e.g.:

```ruby
let!(:bus) { create :bus }
```

and then dynamically refer to its name later:

```ruby
expect(page).to have_text bus.number
```

so that your tests won't fail if we change the factory later.

spec/features/bus_delete_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+	end
+end
+
+describe 'deleting a bus - master of ceremonies' do 
+	before :each do 
+		create :bus
+		when_current_user_is :master_of_ceremonies
+		visit buses_url
+	end
+	context 'with admin privelege' do
```
dfaulken: `privilege`

spec/features/quiz_score_addition_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+describe 'updating a quiz score - quiz scorer' do 
+	before :each do
+		create :participant
+		when_current_user_is :quiz_scorer
+		visit quiz_scores_url
+	end
+	context 'with a number filled in' do
+		it 'update the quiz score' do 
+			fill_in 'quiz_score_points_achieved', with: '50'
```
dfaulken: Why do it twice?

dfaulken:
Don't forget to run `rubocop` also.

spec/models/circle_check_score_spec.rb
```diff
@@ -1,5 +1,9 @@
 require 'rails_helper'
 
-RSpec.describe CircleCheckScore, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe CircleCheckScore do 
+	let(:participant) { create :participant }
+	it 'score should get correct value' do 
+		check_score = create :circle_check_score, participant: participant, total_defects: 5, defects_found: 4
```
dfaulken: Hmm, this test doesn't make it clear why 40 is the expected result.

spec/models/quiz_score_spec.rb
```diff
@@ -1,5 +1,9 @@
 require 'rails_helper'
 
-RSpec.describe QuizScore, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe QuizScore do 
+	let(:participant) { create :participant }
+	it 'score should have the correct value' do 
+		quiz = create :quiz_score, participant: participant, points_achieved: 50, total_points: 100
+		expect(quiz.score).to eql 25.0
```
dfaulken: Same here - I'm unclear just from reading this why the answer is 25.0.

spec/models/user_spec.rb
```diff
@@ -1,5 +1,11 @@
 require 'rails_helper'
 
-RSpec.describe User, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe User do
+	describe 'simple validations' do
+		it 'prevents users from having the same email' do
+			create :user, name: 'Bob Smith', email: 'bsmith@example.com'
```
dfaulken: Since the names aren't relevant to this test, better to leave them out and let the factory handle them.

.gitignore
```diff
@@ -15,3 +15,8 @@
 /log/*
 !/log/.keep
 /tmp
+
+# Ignore all of coverage, except the .last_run.json
```
dfaulken: Thank you! 😄

spec/features/circle_check_score_addition_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+  context 'with admin privilege' do
+    it 'adds and updates circle check score' do
+      when_current_user_is :admin
+      visit circle_check_scores_url
+      fill_in 'circle_check_score_defects_found', with: '4'
+      click_on 'Save score'
+      expect(page).to have_text 'Score was saved.'
+      input = find_field :circle_check_score_defects_found
+      expect(input.value).to eql '4'
```
dfaulken: Better to break these out into separate feature specs.

spec/features/quiz_score_addition_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  context 'with quiz scorer privilege' do
+    it 'adds and updates the quiz score' do
+      when_current_user_is :quiz_scorer
+      visit quiz_scores_url
+      fill_in 'quiz_score_points_achieved', with: '50'
+      click_on 'Save score'
+      expect(page).to have_text 'Quiz score was saved.'
+      input = find_field :quiz_score_points_achieved
```
dfaulken: Same here - separate addition vs. updating specs would be good.

spec/models/circle_check_score_spec.rb
```diff
@@ -1,5 +1,13 @@
 require 'rails_helper'
 
-RSpec.describe CircleCheckScore, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe CircleCheckScore do
+  let(:participant) { create :participant }
+  describe 'score' do
+  	it 'has the correct value' do
+  	  score = create :circle_check_score, participant: participant
+  	  defect_percentage = score.defects_found / score.total_defects
```
dfaulken: indentation is still wrong here.

spec/models/maneuver_participant_spec.rb
```diff
@@ -1,5 +1,11 @@
 require 'rails_helper'
-
-RSpec.describe ManeuverParticipant, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+=begin
+describe 'creating maneuvers' do 
+	let!(:participant) {create :participant}
+	let!(:maneuvers) {create :maneuver}
+	let!(:obstacles) {create :obstacles}
```
dfaulken: Better to just delete the file rather than add comments.

spec/features/bus_delete_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+describe 'deleting a bus' do
+  context 'with admin privilege' do
+    let!(:bus) { create :bus }
```
dfaulken: This `let` statement can go after the `describe` so that you state it once rather than three times.

spec/features/circle_check_score_update_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  context 'with admin privilege' do 
+    it 'updates circle check score' do 
+      when_current_user_is :admin
+      visit circle_check_scores_url
+      fill_in 'circle_check_score_defects_found', with: '4'
+      click_on 'Save score'
+      expect(page).to have_text 'Score was saved.'
+      input = find_field :circle_check_score_defects_found
+      expect(input.value).to eql '4'
```
dfaulken: How do you know it isn't 4 to start with when you create the CCS on line 6?

spec/features/quiz_score_addition_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      when_current_user_is :admin
+      visit quiz_scores_url
+      fill_in 'quiz_score_points_achieved', with: '-14'
+      click_on 'Save score'
+      expect(page).not_to have_text 'Quiz score was saved.'
+    end
+  end
+
+  context'out of range quiz score' do
+    it 'will not accept positive number greater than 100' do
```
dfaulken: what's the significance of 100?

spec/features/circle_check_score_update_spec.rb
AkivaGreen: On line 6 the CCS creates a score for us that we will update in the test with a different value. Which is the 4, it is filled in on line 14 so we test for a successful or unsuccessful update in the it tests.

dfaulken: Right, but this test could be a false positive if when you create it on line 6, it's initialized as 4. Then your test would pass because you're not actually changing the value.

AkivaGreen: Could it work, if I create it with a value that would be different from the 4? Like per say 3? That way it would know if it updates or not.

dfaulken: Yes, that's what I'd recommend.
Am 24.05.2016 12:16 schrieb "AkivaGreen" <notifications@github.com>:

PR #34, opened 2016-05-24
TDD - Getting Current Test Suit to Pass

AkivaGreen:
Adding code to allow the current test suit to pass


app/controllers/circle_check_scores_controller.rb
```diff
@@ -2,9 +2,15 @@ class CircleCheckScoresController < ApplicationController
...
-    score = CircleCheckScore.create! score_params
-    redirect_to circle_check_scores_path, notice: 'Score was saved.'
-    PrivatePub.publish_to '/scoreboard', score
+    deny_access && return unless current_user.has_role? :circle_check_scorer
+    score = CircleCheckScore.new score_params
+    if score.save
+      redirect_to circle_check_scores_path, notice: 'Score was saved.'
+      PrivatePub.publish_to '/scoreboard', score
+    else
+      flash[:error] = 'Invalid Score'
```
dfaulken: Take a look at what some other controllers do - there's a way to dynamically return even more information that this.

app/models/circle_check_score.rb
```diff
@@ -3,6 +3,7 @@ class CircleCheckScore < ActiveRecord::Base
 
   validates :participant, :total_defects, :defects_found, presence: true
   validates :participant, uniqueness: true
+  validates :defects_found, numericality: {less_than_or_equal_to: :total_defects, greater_than_or_equal_to: 0}
```
dfaulken: Don't forget to run `rubocop` - this line is a little long.

spec/models/onboard_judging_spec.rb
```diff
@@ -3,3 +3,25 @@
 RSpec.describe OnboardJudging, type: :model do
   pending "add some examples to (or delete) #{__FILE__}"
 end
+
+describe OnboardJudging do
+  let!(:onboard_judging) { create :onboard_judging }
+  describe 'set_score' do
+    it 'private method is run' do
```
dfaulken: Well, here you're just duplicating the content of the private method. Let's work tomorrow on a better way of doing this.

spec/models/participant_spec.rb
```diff
@@ -19,4 +20,19 @@
       end
     end
   end
+  describe 'total_score' do
+    it'participant has correct maneuver score' do
+      expected_total = participant.maneuver_score
+      if participant.onboard_judging.present?
```
dfaulken: Here also, this is just duplicating the content of the existing code.

In general, if you find yourself writing `if` in a test, you're not on the right track. In a test, you structure a setup in a certain way and make assertions - there should never be conditional logic in that scenario.

dfaulken:
This is on generally the right track - I'll be in tomorrow morning and we can talk about what some better ways to structure a few of the tests might be.

app/controllers/participants_controller.rb
```diff
@@ -12,6 +12,7 @@ def assign_number
   end
 
   def create
+    deny_access && return unless current_user.has_role? :judge
```
dfaulken: These should be `:master_of_ceremonies`.

app/models/onboard_judging.rb
```diff
@@ -4,6 +4,11 @@ class OnboardJudging < ActiveRecord::Base
   belongs_to :participant
   validates :participant, uniqueness: true
   validates :score, :minutes_elapsed, :seconds_elapsed, presence: true
+  validates :minutes_elapsed, numericality: {
+    greater_than_or_equal_to: 0 }
+  validates :seconds_elapsed, numericality: {
+    less_than_or_equal_to: 59,
```
dfaulken: Simpler would be

```ruby
validates :seconds_elapsed, inclusion: { in: 0..59 }
```

spec/models/maneuver_participant_spec.rb
```diff
@@ -1,4 +1,32 @@
 require 'rails_helper'
-RSpec.describe ManeuverParticipant, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe ManeuverParticipant do
+  describe 'score' do
+    it'score should be 22' do
```
dfaulken: What's the significance of the number 22? It's difficult to tell from reading this test.

spec/models/maneuver_participant_spec.rb
```diff
@@ -1,4 +1,32 @@
...
+      maneuver = create :maneuver_participant,
+                        speed_achieved: true, obstacles_hit: { 1 => 2, 2 => 1 },
+                        distances_achieved: { [1, 2] => 3 }
+      expect(maneuver.score).to be(22)
+    end
+  end
+
+  describe 'previous participant' do
+    it'finds previous participant' do
+      partip_1 = create :maneuver_participant
```
dfaulken: Generally I've been going for `record` as a shorter way of referring to a `ManeuverParticipant`

spec/models/maneuver_participant_spec.rb
```diff
@@ -1,4 +1,32 @@
...
+      partip_2 = create :maneuver_participant
+      my_maneuver = create :maneuver,
```
dfaulken: In particular, there really isn't a me, my, or I anywhere in idiomatic Rails code

spec/models/onboard_judging_spec.rb
```diff
@@ -1,5 +1,76 @@
 require 'rails_helper'
 
-RSpec.describe OnboardJudging, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe OnboardJudging do
+  describe 'score' do
+    context 'negative onboard judging' do
+      let(:onboard_judging) { create :onboard_judging, minutes_elapsed: 11 }
+      it 'score should be negative' do
```
dfaulken: I think more to the point is that it `allows negative scores`, unlike the other models which don't.

dfaulken:
So, a couple of general comments:

1. The factories are looking a little bloated. While we're here, let's pare them down to contain just the needed attributes.
2. It looks like some of this code is venturing outside of the original scope of this pull request, which was writing `app` code to get the current specs to pass. What I see here are some additional tests - which is great, but would have been better as part of a different pull request. This is for two reasons:
  + it's code with a different purpose, so the sense of intent gets a little lost when reading through the code here - it's tough to remember what's going on at times;
  + this is just a generally long pull request - it's nice to keep things as short as possible.

In the future, I'd recommend keeping code which is being written for separate purposes (i.e. getting the current tests to pass vs. writing new tests) in separate feature branches, from which you can open separate pull requests that can be examined, discussed, and merged separately (and probably more quickly overall).

spec/factories/participants.rb
```diff
@@ -4,4 +4,14 @@
     sequence :number
     bus
   end
+
+  trait :total_score do
```
dfaulken: Hmm, what's the purpose of this trait? I'm not clear on that from reading the name.

spec/features/circle_check_score_addition_spec.rb
```diff
@@ -36,3 +36,33 @@
...
+      expect(page).to have_text expected
+    end
+  end
+
+  context 'when out of range circle check score' do
+    it 'will not accept positive number greater than total points' do
+      when_current_user_is :admin
+      visit circle_check_scores_url
+      input = find_field 'circle_check_score_total_defects'
+      out_of_range = input.value.to_i + 1
```
dfaulken: I think what might be simpler here is just to fill in both fields yourself:

```ruby
fill_in 'circle_check_score_defects_found', with: 5
fill_in 'circle_check_score_total_defects', with: 4
```

spec/models/maneuver_participant_spec.rb
```diff
@@ -1,4 +1,76 @@
...
-  pending "add some examples to (or delete) #{__FILE__}"
+describe ManeuverParticipant do
+  describe 'score' do
+    let!(:record) { create :maneuver_participant, :perfect_score }
+    it'should be a perfect score' do
+      expect(record.score).to be(50)
+    end
+
+    it 'should go down by 10 for each reverse direction' do
+      record.update! reversed_direction: 1
```
dfaulken: If you're going to go the updating route, you may as well do

```ruby
expect { record.update reversed_direction: 1}.to change { record.score }.by(-10)
```

etc.

spec/models/participant_spec.rb
```diff
@@ -19,4 +20,19 @@
       end
     end
   end
+  #   describe 'total_score' do
```
dfaulken: May as well remove these comments.

spec/factories/participants.rb
AkivaGreen: This is supposed to be for the new pull request. I will have it commented it out for now. However, I have changed its name to better reflect it's function as 'initialize_total_score' because it will initialize the attributes necessary to have a valid total score.

spec/models/maneuver_participant_spec.rb
```diff
@@ -1,4 +1,76 @@
...
-  pending "add some examples to (or delete) #{__FILE__}"
+describe ManeuverParticipant do
+  describe 'score' do
+    let!(:record) { create :maneuver_participant, :perfect_score }
+    it'should be a perfect score' do
+      expect(record.score).to be(50)
+    end
+
+    it 'should go down by 10 for each reverse direction' do
+      expect { record.update reversed_direction: 1 }.to change {
```
dfaulken: More idiomatic line breaking would be

```ruby
expect { record.update reversed_direction: 1}
  .to change { record.score }
  .by(-10)
```

spec/models/maneuver_participant_spec.rb
dfaulken: :zap:

dfaulken:
Please pull from `origin/master` and resolve the merge conflict - this [should fix your Travis failure](https://travis-ci.org/umts/umasstransit.rodeo/builds/133140086) as well.

Issue #15, opened 2016-04-04
Test coverage


dfaulken:
Currently none. It would be worth shoring this up before next year.

dfaulken:
@AkivaGreen, @mboneil10, and @oreoluver are working on this together.

PR #35, opened 2016-05-26
Onboard Judging Spec

mboneil10:
Created and fixed onboard judging model spec
Addresses #15

### DIFF
--- a/spec/models/participant_spec.rb
+++ b/spec/models/participant_spec.rb
@@ -21,26 +21,26 @@ describe Participant do
     end
   end
   describe 'total score' do
-    it 'should be zero' do
+    it 'should be zero when initialized' do
       expect(participant.total_score).to eql 0
     end
-    it 'should increase by 40 from quiz score' do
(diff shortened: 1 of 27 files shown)


spec/models/participant_spec.rb
```diff
@@ -20,4 +20,27 @@
       end
     end
   end
+  describe 'total score' do
+    it 'should be zero' do
```
dfaulken: It's difficult to know from this description under what conditions it would be 0.

spec/models/participant_spec.rb
```diff
@@ -20,4 +20,27 @@
...
+      expect(participant.total_score).to eql 0
+    end
+    it 'should increase by 40 from quiz score' do
```
dfaulken: I think what this is actually testing is that adding a quiz score updates the total score. It'd be better if the test didn't depend on specific numbers, so something more like:

```ruby
quiz_score = create :quiz_score
expect { participant.update quiz_score: quiz_score }
  .to change { participant.total_score }
  .by(quiz_score.score)
```

Issue #21, opened 2016-04-04
Travis / CodeClimate / Rubocop


dfaulken:

dfaulken:
Determined that @Anbranin would need to be a team owner to set up Travis and CC.

dfaulken:
Looks like pushing to master didn't trigger a Travis build @Anbranin

dfaulken:
Scratch that - you didn't actually push to master

Anbranin:
wat

dfaulken:
👍

PR #37, opened 2016-05-26
Fix Travis YML file, create intentional Rubocop failure

dfaulken:


dfaulken:
CodeClimate didn't run (rel: #21).

PR #45, opened 2016-06-01
Maneuver spec, distance target

mboneil10:
Addresses #15

### DIFF
--- a/spec/models/distance_target_spec.rb
+++ b/spec/models/distance_target_spec.rb
@@ -1,5 +1,26 @@
 require 'rails_helper'
-
-RSpec.describe DistanceTarget, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe DistanceTarget do
+  describe 'interval_type' do
+    context 'when interval type is 0' do
+      it 'returns inches' do
+        target = create :distance_target, intervals: 0
+        expect(target.interval_type).to eql 'inches'
+      end
+    end
+    context 'when interval type is non-zero' do
+      it 'returns marks' do
+        target = create :distance_target, intervals: 1
+        expect(target.interval_type).to eql 'marks'
+      end
+    end
+  end
+  describe 'index' do
+    it 'returns index of distance target' do
+      maneuver = create :maneuver
+      target_1 = create :distance_target, maneuver: maneuver
+      target_2 = create :distance_target, maneuver: maneuver
+      expect(target_1.index).to eql 0
+      expect(target_2.index).to eql 1
+    end
+  end
 end
--- a/spec/models/distance_target_spec.rb
+++ b/spec/models/distance_target_spec.rb
@@ -7,7 +7,7 @@ describe DistanceTarget do
         expect(target.interval_type).to eql 'inches'
       end
     end
-    context 'when interval type is non-zero' do
+    context 'when interval attribute is non-zero' do
       it 'returns marks' do
         target = create :distance_target, intervals: 1
         expect(target.interval_type).to eql 'marks'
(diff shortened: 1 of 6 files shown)


spec/models/distance_target_spec.rb
```diff
@@ -1,5 +1,26 @@
 require 'rails_helper'
-
-RSpec.describe DistanceTarget, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe DistanceTarget do
+  describe 'interval_type' do
+    context 'when interval type is 0' do
```
dfaulken: interval type? I'm not clear what that means

spec/models/maneuver_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  describe 'previous participant' do
+    let!(:maneuver) { create :maneuver }
+    let!(:record_1) { create :maneuver_participant, maneuver: maneuver }
+    let!(:record_2) { create :maneuver_participant, maneuver: maneuver }
+    it 'finds previous participant' do
+      actual = record_2.participant.number
+      expect(maneuver.previous_participant(actual)).to eql record_1.participant
+    end
+    it 'finds previous participant with no argument' do
+      expect(maneuver.previous_participant.id).to be record_2.participant.id
```
dfaulken: Better to compare the `Participant` objects than their IDs.

spec/models/maneuver_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(maneuver.image_path).to eql 'maneuvers/Foo.png'
+    end
+  end
+  describe 'grouped_obstacles' do
+    context 'with the same point values' do
+      it 'returns an array of obstacles' do
+        maneuver = create :maneuver
+        obstacle_1 = create :obstacle, maneuver: maneuver
+        obstacle_2 = create :obstacle, maneuver: maneuver
+        expected = { [obstacle_1.point_value, obstacle_1.obstacle_type] =>
```
dfaulken: Unclear from the test why this is the expected output.

spec/models/distance_target_spec.rb
```diff
@@ -1,5 +1,26 @@
...
-  pending "add some examples to (or delete) #{__FILE__}"
+describe DistanceTarget do
+  describe 'interval_type' do
+    context 'when interval attribute is 0' do
+      it 'returns inches' do
+        target = create :distance_target, intervals: 0
+        expect(target.interval_type).to eql 'inches'
+      end
+    end
+    context 'when interval type is non-zero' do
```
dfaulken: Same here

spec/models/maneuver_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(maneuver.image_path).to eql 'maneuvers/Foo.png'
+    end
+  end
+  describe 'grouped_obstacles' do
+    context 'with the same point values' do
+      it 'returns a hash of obstacles' do
+        maneuver = create :maneuver
+        obstacle_1 = create :obstacle, maneuver: maneuver
+        obstacle_2 = create :obstacle, maneuver: maneuver
+        expected = { [obstacle_1.point_value, obstacle_1.obstacle_type] =>
```
dfaulken: If this grouping is because the two obstacles have the same point value and type out of the box, this isn't clear from the specs. If the factory rotates or randomizes the value of those attributes, this test will fail.

spec/models/distance_target_spec.rb
```diff
@@ -1,5 +1,26 @@
...
+    end
+    context 'when interval attribute is non-zero' do
+      it 'returns marks' do
+        target = create :distance_target, intervals: 1
+        expect(target.interval_type).to eql 'marks'
+      end
+    end
+  end
+  describe 'index' do
+    it 'returns index of distance target' do
```
dfaulken: relative to what?

dfaulken:
Closing. Redundant with #46.

PR #46, opened 2016-06-06
Participants model spec

oreoluver:
Addresses #15

### DIFF
--- a/spec/models/distance_target_spec.rb
+++ b/spec/models/distance_target_spec.rb
@@ -1,5 +1,26 @@
 require 'rails_helper'
-
-RSpec.describe DistanceTarget, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+describe DistanceTarget do
+  describe 'interval_type' do
+    context 'when interval type is 0' do
+      it 'returns inches' do
+        target = create :distance_target, intervals: 0
+        expect(target.interval_type).to eql 'inches'
+      end
+    end
+    context 'when interval type is non-zero' do
+      it 'returns marks' do
+        target = create :distance_target, intervals: 1
+        expect(target.interval_type).to eql 'marks'
+      end
+    end
+  end
+  describe 'index' do
+    it 'returns index of distance target' do
+      maneuver = create :maneuver
+      target_1 = create :distance_target, maneuver: maneuver
+      target_2 = create :distance_target, maneuver: maneuver
+      expect(target_1.index).to eql 0
+      expect(target_2.index).to eql 1
+    end
+  end
 end
--- a/spec/models/distance_target_spec.rb
+++ b/spec/models/distance_target_spec.rb
@@ -15,7 +15,7 @@ describe DistanceTarget do
     end
   end
   describe 'index' do
-    it 'returns index of distance target' do
+    it 'returns index of distance target relative to itself' do
       maneuver = create :maneuver
       target_1 = create :distance_target, maneuver: maneuver
       target_2 = create :distance_target, maneuver: maneuver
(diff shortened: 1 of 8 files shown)


spec/models/participant_spec.rb
```diff
@@ -20,27 +20,126 @@
...
+      create :participant
+      p_2 = create :participant
+      expect(Participant.next_number).to eql p_2.number + 1
+    end
+  end
 end
+describe 'scoreboard order' do
+  context 'total score' do
+    it 'sorts participants by total score' do
+      per = create :onboard_judging, :perfect
```
dfaulken: It's a good way to set up the test - although, given that none of your lines of code come even *close* to 80 characters, I would recommend more longer, more descriptive variable names to make it easier to read.

spec/models/participant_spec.rb
```diff
@@ -20,27 +20,126 @@
...
+      expected = [p_1, p_2, p_3]
+      expect(Participant.scoreboard_order(:participant_number)).to eq expected
+    end
+  end
+end
+
+describe 'top_20' do 
+  context'doesnt include last person' do
+    it 'holds 20 people' do
+      21.times { create :participant }
```
dfaulken: Seems a little bit fragile - why in particular is it the last participant you create that should be excluded?

spec/models/participant_spec.rb
```diff
@@ -20,27 +20,126 @@
...
+    end
+    context 'displays name' do
+      it 'shows name' do
+        participant.update! name: 'Person'
+        expect(participant.display_information(:name)).to eql 'Person'
+      end
+    end
+    context 'displays three options' do
+      it 'splits up options' do
+        exp = participant.name.to_s + ' (' + participant.bus.number +
```
dfaulken: These two lines aren't that easy to read.

spec/models/participant_spec.rb
```diff
@@ -20,27 +20,126 @@
...
-    it 'should increase by onboard judging score' do
+    it 'increases by onboard judging score' do
       onboard_judging = create :onboard_judging
       expect { participant.update onboard_judging: onboard_judging }
         .to change { participant.total_score }
         .by(onboard_judging.score)
     end
   end
+  describe 'maneuver_score' do
+    it 'returns the maneuver score' do
```
dfaulken: This isn't really a test of much. I agree that if you just make a participant with no objects they should have a score of 0. Better to create some maneuver_participants and make sure that there score is counted.

spec/models/maneuver_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      maneuver = create :maneuver, name: 'Foo'
+      expect(maneuver.image_path).to eql 'maneuvers/Foo.png'
+    end
+  end
+  describe 'grouped_obstacles' do
+    context 'with the same point values and types' do
+      it 'returns a hash of obstacles' do
+        maneuver = create :maneuver
+        obstacle_1 = create :obstacle, maneuver: maneuver,
+                                       point_value: 4,
```
dfaulken: The specific point_value and obstacle_type don't matter - just that they have the same ones. This is to say

```ruby
obstacle_2 = create :obstacle, maneuver: maneuver,
                               point_value: obstacle_1.point_value,
                               obstacle_type: obstacle_1.obstacle_type
```

works just as well and doesn't imply that the specific data matters.

spec/models/distance_target_spec.rb
```diff
@@ -1,5 +1,26 @@
...
+    end
+    context 'when interval attribute is non-zero' do
+      it 'returns marks' do
+        target = create :distance_target, intervals: 1
+        expect(target.interval_type).to eql 'marks'
+      end
+    end
+  end
+  describe 'index' do
+    it 'returns index of distance target' do
```
dfaulken: relative to what?

dfaulken:
Looks like this pull request was made on top of #45's changes. I moved my comments from #45 over to here.

PR #48, opened 2016-06-07
Controller specs

AkivaGreen:
Addresses #15


db/schema.rb
```diff
@@ -142,4 +142,22 @@
 
   add_index "versions", ["item_type", "item_id"], name: "index_versions_on_item_type_and_item_id", using: :btree
 
+  create_table "views", force: :cascade do |t|
```
dfaulken: A views table? Where'd this come from?

dfaulken: Oh I know, you must've run `rails generate devise views` instead of `rails generate devise:views`. Please remove these changes.

db/seeds.rb
```diff
@@ -147,4 +147,8 @@
                             email: 'dave@example.com',
                             password: 'password',
                             admin: true
+  FactoryGirl.create :user, name: 'Juice Box',
```
dfaulken: Well okay, it'd be nicer if it were at least a *plausible* name for a person.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe Admin::UsersController do
+  include Devise::TestHelpers
+  describe 'update' do
+    context 'bad PUT #UPDATE' do
+      it 'will not update name or email' do
+        request.env['HTTP_REFERER'] = admin_users_url
+        user = create :user, admin: true
+        sign_in user
+        attrs = FactoryGirl.attributes_for(:user, name: '')
```
dfaulken: It should be sufficient just to have

```ruby
attrs = { name: '' }
```

right?

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+        put :update, id: user, user: attrs
+        user.reload
+        expected = ["Name can't be blank"]
+        expect(response).to redirect_to(:admin_users)
```
dfaulken: Wait, this works? Shouldn't it be

```ruby
expect(response).to redirect_to admin_users_url
```

spec/controllers/faye_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+describe FayeController do
+  include Devise::TestHelpers
```
dfaulken: It'd be better to include this at the config level, in either `spec_helper` or `rails_helper`.

spec/controllers/faye_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  before :each do
+    user = create :user, admin: true
```
dfaulken: See, this is why I wrote a helper method for doing this in one line. Wouldn't it be nicer to just write

```ruby
when_current_user_is :admin
```

Could we update that spec_helper method to use the devise test helper?

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+
+describe 'assigning roles' do
+  context 'with admin privilege' do
+    it 'assigns one role to user' do
+      admin = create :user, admin: true
+      login_as admin
```
dfaulken: What's the difference between `sign_in` and `login_as`?

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+
+describe 'assigning roles' do
```
dfaulken: Good name - can we apply it to the file as well?

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      visit new_user_registration_url
+      fill_in 'user_name', with: 'Foo Bar'
+      fill_in 'user_password', with: 'password'
+      fill_in 'user_password_confirmation', with: 'password'
+      expect do
+        within('.actions') do
+          click_on 'Sign up'
+        end
+      end
+        .to change { User.count }
```
dfaulken: Ouch - Rubocop was okay with this? An `end`, a new line, and then a `.`?

I would prefer

```ruby
expect do
  within('.actions'){ click_on 'Sign up' }
end.to change { User.count }.by 0
```

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+  context 'after signup' do
+    it 'allows the user to only view scoreboard' do
+      user = create :user
+      create :participant
+      login_as user
+      visit root_path
+      within 'nav' do
+        expect(page).to have_text 'Scoreboard'
```
dfaulken: Does that do anything? When you say `within` some div, does `page` then only contain the contents of that div, or is `page` still the whole page?

spec/models/obstacle_spec.rb
```diff
@@ -1,5 +0,0 @@
-require 'rails_helper'
```
dfaulken: 👍

db/schema.rb
AkivaGreen: should I run rails destroy devise views? then rails generate devise:views?

dfaulken: Should be sufficient to remove these lines from the schema file.

spec/controllers/admin/users_controller_spec.rb
AkivaGreen: Yes during a binding.pry session and stackoverflow we found that passing :admin_users will automatically convert it to admin_users_url. However,  I will still change it so it reflects ruby idiomatically.

spec/controllers/faye_controller_spec.rb
AkivaGreen: Apologies, there were times in tests where we needed to have a variable for a user that is admin.

spec/features/roles_spec.rb
AkivaGreen: Apologies, these can be changed in to when_current_user_is :admin instead.

spec/features/signup_page_spec.rb
AkivaGreen: It seems that rubocop changed it to this for us. It originally used curly braces.

spec/features/roles_spec.rb
dfaulken: Sure, that works too.

spec/controllers/admin/users_controller_spec.rb
dfaulken: Weird, it must be because you're in that controller spec, it matches symbols to controller actions (the controller does this too).

spec/features/signup_page_spec.rb
AkivaGreen: That is true, this test would then give something like a false positive. I will revise the test, what we are essentially doing is testing that certain links don't appear if you are a person that doesn't have the correct privilege.

spec/controllers/faye_controller_spec.rb
AkivaGreen: It seems that when I use the when_current_user_is :admin method the test fails. However, when I leave it as is the test passes?

spec/features/roles_spec.rb
AkivaGreen: Also when I use sign_in I get a undefined method error in comparision to using login_as.

spec/controllers/faye_controller_spec.rb
dfaulken: Right, because [the spec helper method](https://github.com/umts/umasstransit.rodeo/commit/aed0c165cf92b57b8ed74c4948a12348fe09aaab#diff-93830fa29d616f7c87903d08b5b1b29aR36) uses `login_as`. Hence my question:

> What's the difference between `sign_in` and `login_as`?

spec/features/signup_page_spec.rb
dfaulken: Sure - it'd be nice to find a way to make assertions about the content within a particular element.

spec/controllers/faye_controller_spec.rb
AkivaGreen: sign_in is used by devise::test_helper and login_as is used by spec helper.

spec/features/roles_spec.rb
AkivaGreen: sign_in is used by devise and can't be used in feature tests. However, login_as is used by warden which can be used in feature tests.

spec/features/signup_page_spec.rb
dfaulken: Yet another demonstration that style guides are just guides.

spec/features/roles_spec.rb
dfaulken: Sounds like if we leave the spec_helper method as is and use it everywhere we should be fine.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.54
+    "covered_percent": 8.71
```
dfaulken: Whoops

db/schema.rb
```diff
@@ -142,4 +142,7 @@
 
   add_index "versions", ["item_type", "item_id"], name: "index_versions_on_item_type_and_item_id", using: :btree
 
+  add_index "views", ["email"], name: "index_views_on_email", unique: true, using: :btree
```
dfaulken: Whoops

spec/features/user_delete_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      admin = create :user, name: 'Bob', admin: true
+      login_as admin
+      visit admin_users_url
+      click_on 'Remove', match: :first
+      expect(page).to have_text 'User has been removed.'
+    end
+  end
+  context 'with master of ceremonies privilege' do
+    it 'will not delete a user' do
+      create :user, name: 'Adam'
```
dfaulken: Is the name `Adam` relevant?

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+  context 'a valid user' do
+    it 'creates the user' do
+      visit new_user_registration_url
+      fill_in 'user_name', with: 'Foo Bar'
+      fill_in 'user_email', with: 'foo@valid.com'
+      fill_in 'user_password', with: 'password'
+      fill_in 'user_password_confirmation', with: 'password'
+      expect do
+        within('.actions'){ click_on 'Sign up' }
```
dfaulken: CodeClimate doesn't like the spacing around the curly braces here.

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+  context 'after signup' do
+    it 'allows the user to only view scoreboard' do
+      user = create :user
+      create :participant
+      login_as user
+      visit root_path
+      expect do
+        within('nav').to have_text 'Scoreboard'
```
dfaulken: Uh... what?

dfaulken: This isn't an `expect`ation and is probably not doing what you think it does.

dfaulken: Sorry, missed the `expect do`. Let's see if we can find a better way to do this.

spec/features/user_delete_spec.rb
AkivaGreen: We had a issue with selecting the correct person to remove. If we removed the admin, the page would redirect to the log in page. Since the roles page is sorted by name we decided to give names to the users we create in case there is a change in the factory with the defaults.

spec/features/signup_page_spec.rb
dfaulken: So it looks like from inside a `within` call, you can do `current_scope` to get the element you're within, and then make assertions about it.

```ruby
page.has_text? 'designed and hosted by' # => true
within 'nav' do
  current_scope.has_text? 'designed and hosted by' # => false
end
```

I think that reads nicely - let's use this to do these tests.

spec/features/user_delete_spec.rb
dfaulken: Right, the user needs to be present, sure. But their name shouldn't matter for your test. If it does, then you need a more specific way of running the test.

spec/features/signup_page_spec.rb
AkivaGreen: I was able to use

expect(page).to have_selector('nav', text: 'Scoreboard')
and
expect(page).not_to have_selector('nav', text: 'Maneuver')

They work perfectly, however, would your method be preferred?

spec/controllers/faye_controller_spec.rb
AkivaGreen: I tried adding it to the spec_helper and I get and uninitialized constant error. When I add it outside config just as is only one test fails. However, if I add it using config like such:

config.include Devise::TestHelpers

I get almost all tests failing.

spec/features/signup_page_spec.rb
dfaulken: I think so - cuts down on the repetition.

spec/controllers/faye_controller_spec.rb
dfaulken: Let's move this to Slack and look into what the test failure tells us.

mboneil10: Got help from Brian and Matt. Arta and I have it working.

spec/controllers/faye_controller_spec.rb
werebus: >  Could we update that spec_helper method to use the devise test helper?

From the Devise readme:

>  These helpers are not going to work for integration tests driven by Capybara or Webrat. They are meant to be used with functional tests only. Instead, fill in the form or explicitly set the user in session;

dfaulken: Ah, I see.

werebus: Devise `master` has `Devise::Test::IntegrationHelpers`, but don't know when there will be a gem release.

app/views/maneuver_participants/show.haml
```diff
@@ -14,7 +14,7 @@
         = number_field_tag "obstacle_#{obstacles.first.id}", 
                             @record.obstacles_hit[point_value] || 0,
                             min: 0, max: obstacles.count
-        %button.btn.btn-primary.increment{type: :button,
+        %button.btn.btn-primary.increment{type: :button, disabled: true,
```
dfaulken: Why disabled by default?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.54
+    "covered_percent": 7.37
```
dfaulken: Whoops

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+require 'spec_helper'
```
dfaulken: Shouldn't need to require both, right?

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe Admin::UsersController, type: :controller do
```
dfaulken: We infer spec type from file location - shouldn't need to tag this with spec type

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+  context 'after signup' do
+    it 'allows the user to only view scoreboard' do
+      user = create :user
+      create :participant
+      login_as user
+      visit root_path
+      within 'nav' do
+        current_scope.has_text? 'Scoreboard'
```
dfaulken: These aren't expectations - should be

```ruby
expect(current_scope).to have_text 'Scoreboard'
```

etc.

spec/features/user_delete_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe 'deleting a user' do
+  context 'with admin privilege' do
+    it 'deletes user' do
+      create :user
+      admin = create :user, admin: true
+      login_as admin
+      visit admin_users_url
+      click_on 'Remove', match: :first
```
dfaulken: Cool, didn't know you could do that.

spec/spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 require 'codeclimate-test-reporter'
 require 'factory_girl_rails'
 require 'simplecov'
+require 'rails_helper'
```
dfaulken: Changing the load order - interesting fix. So does this mean that all of our tests should

```ruby
require 'spec_helper'
```

at the top now?

spec/rails_helper.rb
```diff
@@ -1,13 +1,13 @@
...
 require 'factory_girl_rails'
-require 'spec_helper'
 require File.expand_path('../../config/environment', __FILE__)
 require 'rspec/rails'
 require 'capybara/rails'
 require 'rack_session_access/capybara'
 
 ActiveRecord::Migration.maintain_test_schema!
 RSpec.configure do |config|
+  config.include Devise::TestHelpers, type: :controller
```
dfaulken: Nice, didn't know you could limit inclusions like that. So does this mean we should do the opposite with the Warden test helpers?

```ruby
config.include Warden::Test::Helpers, type: :feature
```

db/schema.rb
dfaulken: Travis build won't pass until this indexing code is removed, since the schema doesn't define the `views` table

dfaulken:
Closing in favor of #53.

Issue #17, opened 2016-04-04
User management interface


dfaulken:
Currently giving people logins is done via the Rails console. It'd be nice for people (besides us) to be able to manage logins for everyone.

PR #50, opened 2016-06-09
Dropdown Menu for Logout, Edit Account

AkivaGreen:
A dropdown menu from the name of current user that has the buttons for edit account and logout in a dropdown menu.
Addresses #17


app/views/layouts/_navbar.haml
```diff
@@ -25,6 +25,7 @@
       .dropdown
         %button.btn.navbar-btn.dropbtn
           = current_user.name
+          &blacktriangledown;
```
dfaulken: I wonder if there's a bootstrap-ier way to do this

dfaulken: Something more like [this](http://getbootstrap.com/components/#dropdowns-example)?

PR #53, opened 2016-06-13
Controller specs2

oreoluver:
Addresses #45


app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  private
+
+  def find_user
+    @user = User.find_by id: params.require(:id)
+  end
+
+  def user_params
+    params.require(:user).permit :name, :email, :admin,
```
dfaulken: Why `:name` and `:email`? Name isn't submitted in the form, just shown - and email isn't there at all.

app/views/devise/shared/_links.html.haml
```diff
@@ -0,0 +1,19 @@
+- if controller_name != 'sessions'
```
dfaulken: Did you edit this view? Does it need to be in `app/views/`?

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 97.54
+    "covered_percent": 95.41
```
dfaulken: Hmm, why's this?

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      it 'will not update name or email' do
+        request.env['HTTP_REFERER'] = admin_users_url
+        user = create :user, admin: true
+        sign_in user
+        attrs = { name: '' }
+        put :update, id: user, user: attrs
+        user.reload
+        expected = ["Name can't be blank"]
+        expect(response).to redirect_to admin_users_url
+        expect(flash[:errors]).to match(expected)
```
dfaulken: why `match`?

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+  context 'after signup' do
+    it 'allows the user to only view scoreboard' do
+      user = create :user
+      create :participant
+      login_as user
+      visit root_path
+      within 'nav' do
+        expect(current_scope).to have_text 'Scoreboard'
+        expect(current_scope).to_not have_text 'Maneuver'
```
dfaulken: Prefer `not_to` to `to_not`

spec/rails_helper.rb
```diff
@@ -1,13 +1,13 @@
...
 require 'rspec/rails'
 require 'capybara/rails'
 require 'rack_session_access/capybara'
 
 ActiveRecord::Migration.maintain_test_schema!
 RSpec.configure do |config|
+  config.include Devise::TestHelpers, type: :controller
   config.infer_spec_type_from_file_location!
   config.use_transactional_fixtures = true
   config.include Warden::Test::Helpers
```
dfaulken: Same question here: can we include these for only feature tests?

spec/features/user_delete_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
```
dfaulken: Do the load order changes mean that all of our spec files should require `spec_helper` now?

spec/features/signup_page_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        expect(current_scope).to_not have_text 'Circle Check'
+        expect(current_scope).to_not have_text 'Quiz'
+        expect(current_scope).to_not have_text 'Participants'
+        expect(current_scope).to_not have_text 'Buses'
+        expect(current_scope).to_not have_text  'Roles'
```
dfaulken: Too many spaces here

app/controllers/admin/users_controller.rb
oreoluver: if we take out name and email from the user params then we have to delete one of our tests that tests for coverage of the else part of the update method because we don't know how else to make a faulty update

spec/rails_helper.rb
oreoluver: do you want us to remove it from the rails_helper file and put it in only the feature tests that use it?

coverage/.last_run.json
oreoluver: Some of our test files are spread out across different branches so I think when we merge that should solve the coverage problems

app/controllers/admin/users_controller.rb
dfaulken: App logic should never change to accommodate your tests. Try submitting data other than `true` or `false`, which shouldn't be valid for a boolean column.

coverage/.last_run.json
dfaulken: The only way coverage can go down is if you're adding code but not testing it.

spec/rails_helper.rb
dfaulken: https://github.com/umts/umasstransit.rodeo/pull/48#discussion_r66708307

app/views/maneuver_participants/show.haml
```diff
@@ -26,7 +26,7 @@
         = number_field_tag "target_#{target.id}",
           @record.distances_achieved[[target.minimum, target.multiplier]],
           min: 0
-        %button.btn.btn-primary.increment{type: :button,
+        %button.btn.btn-primary.increment{type: :button, disabled: true,
```
dfaulken: Why disabled by default?

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  private
+
+  def find_user
+    @user = User.find_by id: params.require(:id)
+  end
+
+  def user_params
+    params.require(:user).permit :admin,
```
dfaulken: Weird arrangement of these symbols

spec/controllers/faye_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+describe FayeController do
+  before :each do
+    current_user = create :user, :admin
+    sign_in current_user
+  end
+  context 'post #test' do
+    it 'posts to faye test' do
+      expect(PrivatePub).to receive(:publish_to).with('/test', {})
```
dfaulken: Don't need the ending parentheses here (and in some other places too).

app/controllers/admin/users_controller.rb
oreoluver: which ones??

dfaulken:
Looks great - just my outstanding question about whether our tests should be changed to require `spec_helper` instead of `rails_helper`.

spec/features/user_delete_spec.rb
oreoluver: they all require rails helper and they all pass

PR #55, opened 2016-12-01
Optimize scoreboard by fetching objects in advance

### LINKED ISSUE #18: Optimize scoreboard
It'd be pretty easy to serve up the scoreboard as one object instead of one call per maneuver per participant.

dfaulken:
Closes #18.

Also add `.ruby-version` file.


app/controllers/participants_controller.rb
```diff
@@ -67,9 +53,6 @@ def update
     end
   end
 
-  def welcome
```
dfaulken: In Rails 4+, don't need empty controller action. Declaring the route in `config/routes.rb` is sufficient.

PR #57, opened 2016-12-02
Require registrations to be approved

### LINKED ISSUE #56: User management
In 2016 I just added people via the Rails console.

We could allow anyone to register and then admins to approve like I did on [secret-santa](https://github.com/dfaulken/secret-santa/tree/master/app/controllers/users_controller.rb), or @werebus suggested allowing the master of ceremonies to issue invitations.

dfaulken:
Closes #56.


dfaulken:
Actually, CodeClimate, I think you'll find I didn't add that code here.

PR #59, opened 2017-04-24
Spec trans

Anbranin:
Don't merge this
please
it will break everything

### DIFF
--- a/app/controllers/wheelchair_maneuvers_controller.rb
+++ b/app/controllers/wheelchair_maneuvers_controller.rb
@@ -1,0 +1,40 @@
+class WheelchairManeuversController < ApplicationController
+  before_action :find_record, only: %i(show update)
+
+  def create
+    deny_access && return unless current_user.has_role? :judge
+    record = WheelchairManeuver.create! params.require(:wheelchair_maneuver).permit!
+    redirect_to select_participant_wheelchair_maneuvers_path
+    update_scoreboard with: record
+  end
+
+  def new
+    @wheelchair_maneuver = WheelchairManeuver.new
+    @participant = Participant.find_by(number: params.require(:participant))
+  end
+
+  def select_participant
+    @participants = Participant.all
+  end
+
+  def show
+    @wheelchair_maneuver = WheelchairManeuver.find(params.require(:id))
+    @participant = @wheelchair_maneuver.participant
+  end
+
+  def update
+    deny_access && return unless current_user.has_role? :judge
+    number_wrong = params[:checks].values.map(&:to_i).select(&:zero?).count
+    @record.update number_wrong: number_wrong
+    redirect_to select_participant_wheelchair_maneuvers_path,
+                notice: 'Onboard score has been saved.'
+    update_scoreboard with: @record
+  end
+
+  private
+
+  def find_record
+    @record = WheelchairManeuver.find(params.require(:id))
+  end
+
+end
--- a/app/models/maneuver.rb
+++ b/app/models/maneuver.rb
@@ -5,6 +5,9 @@ class Maneuver < ActiveRecord::Base
   has_many :distance_targets
 
   validates :name, :sequence_number, presence: true, uniqueness: true
+  validates_presence_of :reverse_points,
+    if: Proc.new { |m| m.counts_reverses == true },
+    message: 'Reverses have to count for something!'
 
   def grouped_obstacles
     obstacles.group_by { |o| [o.point_value, o.obstacle_type] }
--- a/app/models/wheelchair_maneuver.rb
+++ b/app/models/wheelchair_maneuver.rb
@@ -22,6 +22,7 @@ class WheelchairManeuver < ActiveRecord::Base
     }
   }
   POINT_VALUES = {
+    # points to remove if participant did NOT do the thing
     first_ask_to_touch: 10,
     first_check_brakes_on: 15,
     offer_seatbelt: 15,
@@ -37,7 +38,7 @@ class WheelchairManeuver < ActiveRecord::Base
   def set_score
     score = 200
     POINT_VALUES.each do |key, value|
-      if self.send(key) == true
+      if self.send(key) == false
         score -= value
       end
     end
--- a/app/models/wheelchair_maneuver.rb
+++ b/app/models/wheelchair_maneuver.rb
@@ -1,0 +1,47 @@
+class WheelchairManeuver < ActiveRecord::Base
+  has_paper_trail
+
+  belongs_to :participant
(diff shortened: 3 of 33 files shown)


app/controllers/wheelchair_maneuvers_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def show
+    @wheelchair_maneuver = WheelchairManeuver.find(params.require(:id))
+    @participant = @wheelchair_maneuver.participant
+  end
+
+  def update
+    deny_access && return unless current_user.has_role? :judge
+    number_wrong = params[:checks].values.map(&:to_i).select(&:zero?).count
```
dfaulken: You can just do `.count(&:zero?)`

app/models/maneuver.rb
```diff
@@ -5,6 +5,9 @@ class Maneuver < ActiveRecord::Base
   has_many :distance_targets
 
   validates :name, :sequence_number, presence: true, uniqueness: true
+  validates_presence_of :reverse_points,
+    if: Proc.new { |m| m.counts_reverses == true },
```
dfaulken: Can use compact lambda syntax, and not comparing to `true`

app/models/maneuver.rb
```diff
@@ -5,6 +5,9 @@ class Maneuver < ActiveRecord::Base
...
+    message: 'Reverses have to count for something!'
```
dfaulken: That message still doesn't help me understand what reverse points are

app/models/wheelchair_maneuver.rb
```diff
@@ -0,0 +1,48 @@
+class WheelchairManeuver < ActiveRecord::Base
+  has_paper_trail
+
+  belongs_to :participant
+  validates :participant, uniqueness: true
```
dfaulken: And presence, right?

app/models/wheelchair_maneuver.rb
```diff
@@ -0,0 +1,48 @@
...
+    check_brakes_off: 15,
+    second_ask_to_touch: 10,
+    second_check_brakes_on: 15,
+    ask_if_all_set_on_lift: 10
+  }
+
+  def set_score
+    score = 200
+    POINT_VALUES.each do |key, value|
+      if self.send(key) == false
```
dfaulken: Don't need to compare to `false`

app/models/wheelchair_maneuver.rb
```diff
@@ -0,0 +1,48 @@
...
+    remove_restraints: 50,
+    check_brakes_off: 15,
+    second_ask_to_touch: 10,
+    second_check_brakes_on: 15,
+    ask_if_all_set_on_lift: 10
+  }
+
+  def set_score
+    score = 200
+    POINT_VALUES.each do |key, value|
```
dfaulken: Personally I think `each_pair` is clearer

app/views/wheelchair_maneuvers/new.haml
```diff
@@ -0,0 +1,20 @@
...
+  %hr
+  = form_tag wheelchair_maneuvers_path do
+    = fields_for :wheelchair_maneuver do |f|
+      = f.hidden_field :participant_id, value: @participant.id
+      %h2 Loading:
+      Did the participant:
+      - WheelchairManeuver::CHECKS[:loading].each_pair do |attribute, prompt|
+        .field
+          = f.label prompt
+          = f.check_box "#{attribute}"
```
dfaulken: Why do we string interpolate a string?

db/seeds.rb
```diff
@@ -138,8 +82,12 @@
     CircleCheckScore.create! participant: p, total_defects: 5, defects_found: rand(5)
     QuizScore.create! participant: p, total_points: 100, points_achieved: rand(100)
     OnboardJudging.create! participant: p, minutes_elapsed: (6 + rand(3)), seconds_elapsed: rand(59),
-      missed_turn_signals: 0, missed_flashers: rand(1), missed_horn_sounds: rand(1), times_moved_with_door_open: 0,
-      unannounced_stops: 0, sudden_stops: 0, sudden_starts: 0, abrupt_turns: 0
+      missed_turn_signals: rand(2), sudden_stops: rand(3), sudden_starts: rand(4), abrupt_turns: rand(3)
+    WheelchairManeuver.create! participant: p, first_ask_to_touch: [true,false].sample,
```
dfaulken: We should write a method for `[true, false].sample`

spec/features/recording_maneuver_score_spec.rb
```diff
@@ -27,6 +27,8 @@
     create :maneuver_participant, :perfect_score
   end
   it 'updates the value' do
+    maneuver_participant.maneuver.update_attributes(counts_reverses: true,
```
dfaulken: Why not create it like this?

app/controllers/wheelchair_maneuvers_controller.rb
Anbranin: Or just remove that line because it's for old code and doesn't actually work lol

app/models/maneuver.rb
Anbranin: -_-

app/models/wheelchair_maneuver.rb
Anbranin: So I have to use an unless statement?

dfaulken: You don't have to, but it's simpler.

dfaulken: Also, better descriptors than `key` and `value` would help clarify what's going on here.

Anbranin: Oh, how about `if self.send(key).blank? `

app/views/wheelchair_maneuvers/new.haml
Anbranin: lol

app/models/wheelchair_maneuver.rb
dfaulken: But it's false, not blank.

spec/features/recording_maneuver_score_spec.rb
Anbranin: Couldn't use the 'perfect score' thingy then, made bad things happen

Anbranin: Not sure why need it to be perfect score anyways

Anbranin: Now I know

dfaulken:
We can close this, right?

PR #70, opened 2018-04-26
Same version-pinning, also use the 'st' branch.

werebus:


werebus:
Build failure is fixed by #68

PR #78, opened 2019-04-12
Rails 5 upgrade

werebus:


werebus:
I'll get on that Travis failure now.

app/controllers/admin/users_controller.rb
```diff
@@ -1,45 +1,49 @@
-class Admin::UsersController < ApplicationController
-  before_action :find_user, except: %i(index manage)
+# frozen_string_literal: true
 
-  def approve
-    @user.approve!
-    redirect_to manage_admin_users_path,
-                notice: "#{@user.name} has been approved."
-  end
+module Admin
```
werebus: Github's diff is pretty confused. All I did was change:

```ruby
class Admin::UsersController < ApplicationController
end
```

to

```ruby
module Admin
  class UsersController < ApplicationController
  end
end
```

db/seeds.rb
```diff
@@ -125,8 +125,8 @@
       mp = ManeuverParticipant.new participant: p, maneuver: m, reversed_direction: 0, completed_as_designed: true
       unless m.name.include?('Passenger') || rand(4) == 3
         m.obstacles.order('rand()').take(2).each do |obst|
-          mp.obstacles_hit[obst.point_value] ||= 0
-          mp.obstacles_hit[obst.point_value] += 1
+          mp.obstacles_hit[obst.id] ||= [obst.point_value, 0]
```
werebus: As far as I can tell, this was just an error. The data-structure must have changes since the seeds were last run.

.travis.yml
```diff
@@ -1,11 +1,13 @@
+dist: xenial
```
dfaulken: Curious why we want to lock the distribution. Are we jumping ahead or locking to a past distro?

app/controllers/participants_controller.rb
```diff
@@ -40,6 +41,8 @@ def index
     @buses = Bus.order :number
   end
 
+  def scoreboard; end
```
dfaulken: I thought Rails 5 didn't need empty controller actions.

app/controllers/quiz_scores_controller.rb
```diff
@@ -1,48 +1,20 @@
...
-    if @score.update score_params
-      redirect_to quiz_scores_path, notice: 'Quiz score was saved.'
-      update_scoreboard with: @score
-    else
-      flash[:errors] = @score.errors.full_messages
-      redirect_to :back
-    end
-  end
+class QuizScoresController < ApplicationController
+  include NonManeuverScoreController
```
dfaulken: Very neat, I like this approach

config/secrets.yml
```diff
@@ -1,8 +1,22 @@
...
+# Make sure the secret is at least 30 characters and all random,
+# no regular words or you'll be exposed to dictionary attacks.
+# You can use `rails secret` to generate a secure secret key.
+
+# Make sure the secrets in this file are kept private
+# if you're sharing your code publicly.
+
 development:
-  secret_key_base: 4c987fc2f10296e546c25ee1965d93e6995ed248c679e4e6153c6fa3ab5bb43fd2709322e535dbb911ac2d33d9daf79e0ad0fbd1c33ea6444a9ad201c7a73fb8
+  secret_key_base: be2c4a5ec154182c27c00a3ecc760bc552d768e34437574efd67e80415fda978863899994c9d1357bcdd6ba5ed8f8c42c89cbb87ceee64f1c452d8b540b9cd82
```
dfaulken: Just curious, was this a purposeful rotation? Should we be rotating these periodically on other apps?

.travis.yml
werebus: This is a jump ahead (default is still Trusty). It's related to failures that we were getting related to switching to `utf8mb4`. MySQL <5.7 (well, actually, just InnoDB) has a problem with indexes on `VARCHAR(255)` (`string`) columns and this character set.

If it were _our_ MySQL server, there's a config option to allow "large indices", but Xenial has MySQL 5.7 so this is the easiest way to sidestep the issue.

app/controllers/participants_controller.rb
werebus: It doesn't. But Rubocop doesn't like the presence of actions in `before_action`s that may or may not exist, and I _think_ I agree with it. Because the controller does have a non-default effect on the request (in the form of its `before_action`s)

Seems a little like needless cruft, but is also a bit of a breadcrumb for future developers, perhaps?

dfaulken: Sure, seems reasonable - though I would suggest that a comment to that effect would be a bigger breadcrumb.

config/secrets.yml
werebus: Nope, for some reason the `rails app:upgrade` task replaced them, and it also doesn't matter in the case where the app has been down for 11 months, so I just shrugged.

PR #82, opened 2019-04-18
Wanted a few of these configuration values

werebus:
For other tests I'm writing.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,32 +2,26 @@
 
 require 'factory_bot'
 require 'simplecov'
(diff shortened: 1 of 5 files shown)


spec/spec_helper.rb
```diff
@@ -2,14 +2,16 @@
 
 require 'factory_bot'
 require 'simplecov'
-Dir[Rails.root.join("spec/support/**/*.rb")].each {|f| require f}
+
+Dir[Pathname(__dir__).join('support', '**', '*.rb')].each { |f| require f }
```
Anbranin: wassthis

werebus: Because it's in the `spec_helper`, `Rails` isn't a thing (yet)

PR #85, opened 2019-04-22
Made myself a helper for all those +/- number fields

werebus:


app/helpers/scoring_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def increment_button(target_field, value, type, options)
+    types = { :+ => ['plus', :max], :- => ['minus', :min] }
+    disabled = (options[types[type][1]] && value == options[types[type][1]])
+
+    button_tag types,
+               class: 'btn btn-primary increment',
+               type: :button,
+               data: { field: sanitize_to_id(target_field),
```
Anbranin: API dock is down and info on what `sanitize_to_id` is is scarce--what does it do?

werebus: That gets you e.g. from `onboard_judging[minutes_elapsed]` to `onboard_judging_minutes_elapsed`

That is, from the `name` of the field to the `id` of the field.

Anbranin: oh cool

lib/app_form_builder.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AppFormBuilder < ActionView::Helpers::FormBuilder
```
Anbranin: The fact that you could do this and then specify a `builder` for the form is new to me. Sweet!

werebus: Yeah. You can also, in your app config do something like

```ruby
ActionView::Base.default_form_builder = AppFormBuilder
```

In the past, I've used gems for Bootstrap or Foundation that do this behind the scenes.

PR #90, opened 2019-04-24
Hidden fields to make the check_box_tags work

### LINKED ISSUE #88: Check-boxes in scoring don't update true-to-false
Because they're `checkbox_tag`s, they don't have the Rails magic `false` hidden fields.

That is, by default in HTML/browsers, un-checked check boxes don't get put in the params. So, once an object is saved with a `true` boolean field, it can't be updated to `false`.

werebus:
Closes #88


werebus:
Well, `check_box_tag`s work the same way `<input type="checkbox">` tags have always worked. Blame the W3C, I guess.

PR #91, opened 2019-04-24
Redo style

werebus:
This was pretty much me pulling all the CSS out and allowing Bootstrap to manage nearly everything. A few light touches added back in. I'll put together a screenshot gallery shortly.


werebus:
Mobile:  https://imgur.com/a/zY5GtQn
Desktop:  https://imgur.com/a/llLp72T

app/assets/javascripts/scoring.js
```diff
@@ -16,4 +17,9 @@ $(document).ready(function(){
       if(max && value === parseInt(max)) $(this).prop('disabled', true);
     }
   });
+
+  $('.scoring').on('click', '.pseudo-checkbox', function(){
```
dfaulken: `click` is probably the only event that matters, right? I have this vague sense that I've been burned before by needing to handle some other (maybe mobile?) case for checkboxes...

app/assets/stylesheets/participants.scss
```diff
@@ -0,0 +1,5 @@
+table.participants {
+  input[type=text], input[type=number], select {
+    min-width: 5rem;
```
Anbranin: had to look up what `rem` is, thought it was a typo. Interesting

app/assets/stylesheets/scoreboard.scss
```diff
@@ -0,0 +1,24 @@
+img.scoreboard-sort-loading{
+  height: 30px;
+  padding: 3px;
+}
+
+.scoreboard-content {
+  max-height: 75vh;
```
werebus: You know, maybe I should `calc` a value instead?

app/assets/javascripts/scoring.js
werebus: So, first off, this isn't actually a checkbox (which, I think _would_ make things more complicated). It used to be that there was (on most mobile browsers) a 300ms delay between when the user touched the screen and when the click event fired. This was so that the browser could determine whether you were actually double-tapping-to-zoom and not clicking on something. One work-around to that problem was to listen for `touchstart`, but them you also have to deal with the fact that `click` happens _too_. There are some helper libraries out there that essentially do:

* Listen for `touchstart` or `touchend`
* Prevent default
* Trigger a `click` event

**But**, since around 2015, both mobile Chrome and mobile Safari check for this:

https://github.com/umts/umasstransit.rodeo/pull/91/files#diff-a68281def00c6c6e8ca731b5f1b1727fR5

And if it's there, the assumption is that the user is not double-tapping to zoom since we promise zooming isn't required. So both browsers dispatch the `click` right away.

werebus: Where this technique _really_ falls down is accessibility. I'm not really sure how to make this control keyboard-accessible (it might be possible, though). So, as is, keyboards and assistive devices aren't going to work very well for maneuver scoring.

app/assets/stylesheets/participants.scss
werebus: Interestingly, I'm starting to doubt myself here. I think `em` might be a better choice,

PR #94, opened 2019-04-27
Some icon imporovements

werebus:
1. Replaced the text :heavy_plus_sign:  and :heavy_minus_sign:  with icons
2. Added :arrow_left:  and :arrow_right: icons to next/previous buttons, removed the text from those buttons on small screens.
3. Redid the giant check-boxes: It turns out that Bootstrap kinda has something for a checkbox hidden in a button.


werebus:
Oh, and that new checkbox is keyboard-accessible.

PR #95, opened 2019-04-29
This is a possible other fix to the auto-load/eager-load paths problem.

werebus:
I figure it's actually fine to just set the default form builder for all of them.

There's a _lot_ of debate on the Internet over whether `/lib` is "supposed" to be in `auto_load_paths` or `eager_log_paths`. This works too. Any thoughts?


werebus:
Sold.

PR #96, opened 2019-04-29
Created shared "title" partial

werebus:
Noticed that _most_ pages started with a title of sorts. I pulled that out into the layout/partial. So that that look stays unified.

### DIFF
--- a/app/views/maneuvers/index.haml
+++ b/app/views/maneuvers/index.haml
@@ -1,6 +1,3 @@
-%h1 Which maneuver are you judging?
-%hr
-
 %ul.list-group.maneuvers
   - @maneuvers.each do |maneuver|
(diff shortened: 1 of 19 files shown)


app/views/maneuvers/index.haml
```diff
@@ -1,6 +1,3 @@
-%h1 Which maneuver are you judging?
```
Anbranin: Decide to do away with this question entirely? I don't disagree, just wondered if it was intentional.

werebus: Yeah, the voice seemed pretty different from the other page titles.

dfaulken: Agreed. I did everything one way, but then wanted this page to seem friendlier (you may recall that we had some concerns about less-tech-savvy users being unwilling to use mobile devices at all).

PR #112, opened 2020-03-24
Update Rails 5.0 -> 5.2

werebus:
This started becuase Rails <= 5.2.3 has a vulnerability (CVE-2020-5267), but that, of course required a whole Rails upgrade.  I also:

- Disabled monkey-patching in RSpec
- moved Feature -> System tests


db/migrate/20200325164623_add_unique_indices.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddUniqueIndices < ActiveRecord::Migration[5.2]
```
Anbranin: This I don't get--adding indices makes sense, but why did you have to specify that they are unique?

werebus: This is a suggestion of rubocop-rails

https://github.com/rubocop-hq/rubocop-rails/blob/master/manual/cops_rails.md#railsuniquevalidationwithoutindex

werebus: To be clear, they're not unique among indices, they're indices that ensure uniqueness.

Anbranin:
OHhh okay, that makes more sense.

Issue #61, opened 2017-05-01
Ability to "close" scoring?


werebus:
We had a judge stop by to ask if he could amend an on-board score. It turns he could.

Should the master of ceremonies be able to declare scores "locked" or something?

dfaulken:
Sure, no reason not to. I didn't implement it originally because the site was supposed to be pretty ephemeral.

frothedoatmilk:
This would involve...a lot of work. The _correct_ way to do this would be to implement some high scope model `Competition` which would have a `locked` attribute, and would be the parent of `User`, `Maneuvers`, etc... Not really worth doing for a small feature which could be replaced by the honor code

werebus:
FWIW, it doesn't have to be _quite_ as complicated as that. All `ManeuverParticipant`s, `CircleCheckScore`s, etc. belong to the "current" Roadeo by virtue of existing; we [wipe the database](https://github.com/umts/umasstransit.rodeo/blob/33ac805893e88781c3c6f0a00d83643e957eae89/lib/tasks/reset.rake#L3-L15) every year.

dirschn:
I would think maybe a custom validation tied to some arbitrary thing that checks "if yes, update is valid, if no block it"

werebus:
We actually have a place to tie into already:

https://github.com/umts/umasstransit.rodeo/blob/33ac805893e88781c3c6f0a00d83643e957eae89/app/models/user.rb#L21-L23

If this method returns false, then only Admins and the Master of Ceremonies can access any of the routes that allow entering or modifying scores.

frothedoatmilk:
The problem was figuring out what the "arbitrary thing" was, I was in the mindset of "do it right the first time," but I could just tie in into some existing model, since this needs to persist. Sounds like a job for the user model!

dirschn:
The "arbitrary thing" is what @werebus linked above, that method should be modified to accommodate your needs

frothedoatmilk:
My problem is not the tie in, it's how do I properly persist the setting for locking the code? Another solution I had was to add an attribute to every user, something like `lock_scores`, and if any admin or MC has it set to true, you lock scores? It felt jank initially which is why I didn't implement that way

werebus:
[...]

PR #124, opened 2021-03-15
Nav cleanup

werebus:
Wrote and tested a helper method for navigation items. Also changed the "are they allowed" logic to use `User#role?`.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,7 +16,7 @@ class User < ApplicationRecord
 
   def role?(role)
     admin? ||
-    (master_of_ceremonies? || scoring_enabled?) && send(role)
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -15,7 +15,12 @@ def approve!
   end
 
   def role?(role)
-    admin? || send(role)
+    admin? ||
+    (master_of_ceremonies? || scoring_enabled?) && send(role)
+  end
+
+  def scoring_enabled?
```
werebus, line 22: This is the beginning of #61, too. If this method returns `false`, then only the "Master of Ceremonies" can modify scores. (It effectively takes all other non-admin roles away).

Anbranin (approved):

PR #125, opened 2021-03-16
Results archive

### LINKED ISSUE #73: Scoreboard export
It would be kind-of nice if the scoreboard from year-to-year, could be preserved in a way that doesn't cost us too much money.

We could do that as an entirely static site (hosted at GH Pages or maybe an S3 bucket?). There's some inspiration to be found [here][1].

Could be tied into #61 or not, but it is the case that once the scores are "locked", the site could be static from that point onward.

[1]:  https://nowaker.net/post/ruby-on-rails-a-static-site-generator.html

werebus:
Two parts (with specs for each)

1. A utility class for rendering the scordeboard view
2. Rake tasks for writing that view to a file and for getting a copy of compiled assets

Closes #73

### DIFF
--- a/lib/scoreboard_renderer.rb
+++ b/lib/scoreboard_renderer.rb
@@ -18,9 +18,10 @@ class ScoreboardRenderer < ParticipantsController
   class << self
     def assigns
       {
-        participants: Participant.scoreboard_order(nil),
+        archiving: true,
         can_edit_scores: false,
         maneuvers: Maneuver.order(:sequence_number),
+        participants: Participant.scoreboard_order(nil),
         scores: ManeuverParticipant.scoreboard_grouping
       }
     end
--- a/lib/scoreboard_renderer.rb
+++ b/lib/scoreboard_renderer.rb
@@ -2,14 +2,8 @@
 
 class ScoreboardRenderer < ParticipantsController
   module ScoreboardRendererHelper
-    def javascript_include_tag(*sources)
-      options = sources.extract_options!.merge(debug: false)
-      super(*sources, options)
-    end
-
-    def stylesheet_link_tag(*sources)
-      options = sources.extract_options!.merge(debug: false)
-      super(*sources, options)
+    def debug_assets
+      false
     end
 
     def user_signed_in?
@@ -17,6 +11,7 @@ class ScoreboardRenderer < ParticipantsController
     end
   end
   helper ScoreboardRendererHelper
+
   skip_forgery_protection
 
   class << self
--- a/lib/scoreboard_renderer.rb
+++ b/lib/scoreboard_renderer.rb
@@ -1,0 +1,36 @@
+class ScoreboardRenderer < ParticipantsController
+  module ScoreboardRendererHelper
+    def javascript_include_tag(*sources)
+      options = sources.extract_options!.merge(debug: false)
+      super(*sources, options)
+    end
+
+    def stylesheet_link_tag(*sources)
+      options = sources.extract_options!.merge(debug: false)
+      super(*sources, options)
+    end
+  end
+
+  helper ScoreboardRendererHelper
+
+  class << self
+    def assigns
+      {
+        participants: Participant.scoreboard_order(nil),
+        can_edit_scores: false,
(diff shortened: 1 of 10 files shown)


werebus:
On the 1 codeclimate issue: new Rubocop (#123) advises exactly the opposite.

lib/scoreboard_renderer.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class ScoreboardRenderer < ParticipantsController
+  module ScoreboardRendererHelper
+    def javascript_include_tag(*sources)
+      options = sources.extract_options!.merge(debug: false)
```
Anbranin, line 6: Why?

spec/lib/scoreboard_renderer_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      expect(stylesheets).not_to include(match('debug'))
+    end
+
+    it 'has non-debug javascript' do
+      js = css['script[src]'].map { |s| s['src'] }
+      expect(js).not_to include(match('debug'))
+    end
+
+    it 'has no navigation buttons' do
+      expect(css['nav #mainnav a']).to be_empty
```
Anbranin, line 33: Will this pass if there's no CSS that even matches `nav #mainnav a` or will it throw up?

werebus: If there were CSS that matched `nav #mainnav a`, then it wouldn't be empty.

Anbranin: What I'm getting at is that what if I come along and decide that i'm going to change `#mainnav` to `#somethingelse` or we do a bootstrap update and `nav` isn't a thing?  This test would erroneously pass, right? I suppose it's not the biggest deal.

lib/scoreboard_renderer.rb
Anbranin: Oh gosh, okay, you want the assets to _always_, even in development, be precompiled into application.whatever.

lib/scoreboard_renderer.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def render
+      super(:scoreboard, assigns: assigns)
+    end
+  end
+
+  def flash
+    {}
+  end
```
Anbranin, line 40: Why do you need this method? JW

werebus: I.... don't? Hmmm....

lib/scoreboard_renderer.rb
werebus: Well, the point is that they _are_ compiled as such, and I want the HTML to point to files that exist.

spec/lib/scoreboard_renderer_spec.rb
werebus: There's another test for the existence of an `h1` in the nav. I can leverage that a bit to help guard against that, I think.

lib/tasks/archive.rake
```diff
@@ -0,0 +1,28 @@
...
+
+namespace :roadeo do
+  desc 'Create a static archive of the Roadeo scoreboard'
+  task archive: [:environment, 'archive:index', 'archive:assets']
+
+  namespace :archive do
+    desc 'Archive the HTML scoreboard to `/archive/index.html`'
+    task index: ARCHIVE.join('index.html')
+
+    file ARCHIVE.join('index.html') => :environment do |t|
```
Anbranin, line 14: I'm afraid to ask this question but I really don't think I'm going to find the answer myself, this time. It seems like this can only make a file called `index.html`. Wouldn't that preclude making one for every year? Do we just want one archive file ever?

werebus: My idea was that the "archive" for a given year is a self-contained "site", yes.

You _can_ do otherwise, though; it just dumps everything in the `archive/` directory. Since you're going to have to move those files.... somewhere in order to host them as a website, you could also rename them in the process (even manually edit the HTML if you need to).

Anbranin (approved):

PR #126, opened 2021-03-16
Some specs for the more obscure view helpers

werebus:
Also beefed up the documentation on these two methods.


werebus:
2 CodeClimate issues are both line-length of comments. They'll go away with the new version in #123

Anbranin (approved):

PR #128, opened 2021-04-02
Installed and ran rubocop-rspec

werebus:
Whew, this one ended up being way more work that I expected.


app/models/distance_target.rb
```diff
@@ -7,10 +7,6 @@ class DistanceTarget < ApplicationRecord
             :multiplier, :maneuver, :name, :minimum,
             presence: true
 
-  def index
-    maneuver.distance_targets.order(:id).index self
-  end
```
werebus, line 12: This method was unused and I opted to remove it rather than update its spec

.rubocop.yml
```diff
@@ -22,5 +24,13 @@ Metrics/ClassLength:
 Metrics/MethodLength:
   Max: 20
 
+RSpec/DescribeClass:
+  Exclude:
+    - spec/tasks/**/*
+    - spec/system/**/*
+
+RSpec/MessageSpies:
+  EnforcedStyle: receive
```
werebus, line 33: I actually like the alternative a bit better, but it's a lot of work. Just FYI, though:

Instead of something like this

```ruby
expect(:my_thing).to receive(:a_method)
:my_thing.a_method
```

You can do this:

```ruby
thing = spy(:my_thing)
thing.a_method
expect(thing).to have_received(:a_method)
```

The reason I like this is that it gets rid of the need to reverse the setup/expectation order when dealing with mocks. You still setup the test, perform the action, then assert the expectation — same order no matter what.

app/views/participants/_table.haml
```diff
@@ -8,23 +8,25 @@
...
           %td= f.number_field :number, class: 'form-control'
           %td= f.collection_select :bus_id, @buses, :id, :number,
             {include_blank: true}, {class: 'form-control'}
           %td= f.submit 'Add', class: 'btn btn-primary'
           %td
       - @participants.each do |participant|
-        %tr
-          = form_for participant do |f|
+        %tr{ data: { participant_id: participant.id } }
+          = form_with model: participant, namespace: participant.id, local: true do |f|
```
werebus, line 21: View fixes make the tests a bit easier, but also fix the fact that every participant name box on this page had the same id

Anbranin (approved):

PR #129, opened 2021-04-23
Remove deprecated codeclimate-test-reporter gem 

werebus:


.travis.yml
```diff
@@ -2,19 +2,32 @@ dist: focal
...
   - yes | gem update --system --force
   - gem install bundler
-before_script:
-  - cp config/database.yml.example config/database.yml
-  - bundle exec rake db:reset
   - nvm install 'lts/*'
   - nvm use 'lts/*'
+install:
+  - bundle config set --local deployment 'true'
+  - bundle install --jobs=3 --retry=3
```
jforgue, line 16: What's the max limit for concurrent jobs on Travis? Could we go higher than 3?(not that we need to)

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 98.34
+    "line": 98.99
```
jforgue, line 3: "line"? That's new, is it supposed to be like that?

jforgue (commented):
Few questions, all else looks good

.travis.yml
werebus: It has to do with the number of CPU cores, setting this higher than N-1 won't offer any improvement. I don't know for sure what the limit is, but this is what [Travis does in their default config](https://docs.travis-ci.com/user/languages/ruby/#dependency-management).

coverage/.last_run.json
werebus: Yes, SimpleCov 0.18 (released in January, 2020) added "branch coverage". The type of coverage we _have_ been using is "line coverage" they renamed it to avoid ambiguity. The test reporter gem was holding back our version of SimpleCov which is why this change showed up here.

jforgue (approved):

PR #140, opened 2022-02-15
Bundle update / Rails update

werebus:
This is primarily for the security update in Rails 6.1.4.6

But, I'm much more cavalier with this app since its not deployed anywhere at the moment. Let's just update everything.


Anbranin (approved):
Whee

PR #146, opened 2022-04-14
Install Capistrano Git plugin

werebus:
Git was the default once upon a time, but depending on that is deprecated.

### DIFF
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-remote_user = Net::SSH::Config.for('umasstransit.rodeo')[:user] || ENV['USER']
-server 'umasstransit.rodeo',
-       roles: %w[app db web],
-       user: remote_user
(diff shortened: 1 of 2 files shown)


config/deploy/production.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-remote_user = Net::SSH::Config.for('umasstransit.rodeo')[:user] || ENV['USER']
-server 'umasstransit.rodeo',
-       roles: %w[app db web],
-       user: remote_user
```
werebus, line 6: Newer Capistrano can figure this out from SSH without being forced to.

config/deploy/production.rb
```diff
@@ -1,7 +1,5 @@
...
-set :tmp_dir, "/tmp/#{remote_user}"
```
werebus, line 7: This was a workaround for different users trying to use the same git wrapper. Capistrano now makes a new one every t.ime

Anbranin (approved):

PR #147, opened 2022-04-14
Remove "password generator" script

werebus:
This isn't used anymore, but it does hint at passwords that might be used by... someone?


dirschn:
Is there a separate database to load for this repo?

Anbranin (approved):

werebus:
> Is there a separate database to load for this repo?

You're asking about development data, I think? The way we do things on round-three — where we take regular obfuscated dumps of the production database — is kinda the exception (both at UMTS and elsewhere in the industry). The most common way to populate Rails development data is using a "[seeds file](https://github.com/umts/umasstransit.rodeo/blob/master/db/seeds.rb)".

Rails has a task, `rails db:seed`, that runs that file. That file can technically have anything in it, but the understanding is that it will create database objects...somehow. There is also `rails db:setup`, which creates the DB, loads the schema file, and _then_ runs the seed file; `rails db:prepare`, which is like `db:setup` if the database doesn't exist yet, like `db:migrate` if it does; and `rails db:reset` which first drops the DB, then does what `db:setup` does.

Finally, we're going to try to get better at surfacing this, but the [`bin/setup`](https://github.com/umts/umasstransit.rodeo/blob/master/bin/setup) script _should_ take care of (nearly) all this stuff for you.

dirschn:
Okay, cool, thank you for the info. I'll try all that, i attempted running the db commands but ran into some issues but with the new info it should go smoother.. hopefully! I won't get a chance to do it til a bit later so if you could keep the branch open that'd be great. Though i guess there's other PRs and All this was was deleting a file. 

In any event, thanks, and I'll try it out later

PR #150, opened 2022-04-15
include safari for full screen usage

### LINKED ISSUE #149: full screen button on scoreboard inop
I've clicked it a few times both in and out of full screen mode and it seems as though it's entirely inop

edit: in safari

dirschn:
closes #149 

noticed the full screen button didn't work on safari, so I added safari functionality

### DIFF
--- a/app/assets/javascripts/scoreboard.js
+++ b/app/assets/javascripts/scoreboard.js
@@ -3,6 +3,8 @@ $(document).ready(function(){
     var scoreboard = $('.scoreboard-content').get(0);
     if (scoreboard.requestFullscreen) {
       scoreboard.requestFullscreen();
+    } else if (scoreboard.webkitRequestFullscreen) { /* Safari */
+      scoreboard.webkitRequestFullscreen();
     }
   });
 


werebus (approved):
https://www.safari-is-the-new-ie.com/

this doesn't really do any harm, but I will note:

1. In general, I don't have too much interest in going out of my way to support Safari. They're over 5 years behind every other browser on this one, for example.
2. The full-screen functionality really only has to work on one computer: the scoreboard display at the Roadeo.

dirschn:
Yeah, i figured that was all the case, but it also seemed funny to me that
for the machines we do our development on we can't even test that feature
using the native browser (which is basically the whole point of the
website). But yeah, i guess it's pretty easy to switch browsers

werebus:
First off, I want to clarify that I approved this PR. Sometimes my grumbling asides can come off as denials.

For what it's worth, "we" don't all develop on Macs. On Linux, I can install any browser but Safari. On Windows, I can install any browser but Safari, and on MacOS I can install any browser and also Safari. As long as that's the case, _I_ won't be developing for it.

However, that's not to say "we" shouldn't try. In particular, this site is aimed at smartphones perhaps more than many of our projects, and  you can't install anything but Safari on an iPhone. It just happens that you were working on pretty much the only feature on the site that's designed with a giant TV in mind rather than a phone.

dirschn:
> First off, I want to clarify that I approved this PR. Sometimes my grumbling asides can come off as denials.

Yeah, that's also my bad- I was only looking at my email on my phone I didn't actually check the website.

> For what it's worth, "we" don't all develop on Macs.

Good point, I made the assumption because that's what I was handed when I started.

>  It just happens that you were working on pretty much the only feature on the site that's designed with a giant TV

Right, I just thought it was possible that whoever runs it uses a mac and then tries using Safari, only to find out they can't actually make it full screen. Adam popped into our roadeo meeting though and mentioned how the tv should have a built in browser, so perhaps it's a non-issue.

PR #152, opened 2022-04-15
fat-finger proof

### LINKED ISSUE #151: fat-finger proof the `increment_button`
I may have a slight obsession with giving elements space but I think in this case it's justified... more space between the buttons to prevent accidental pressing of the wrong button

dirschn:
closes #151

### DIFF
--- a/app/helpers/scoring_helper.rb
+++ b/app/helpers/scoring_helper.rb
@@ -34,7 +34,7 @@ module ScoringHelper
     }
     disabled = (options[types[type][1]] && value == options[types[type][1]])
 
-    button_tag(class: 'btn btn-primary increment input-group-append',
+    button_tag(class: 'btn btn-primary increment input-group-append mx-3',
                type: :button,
                data: { field: sanitize_to_id(target_field),
                        type: types[type][0] },


werebus:
I... don't like it.

The spacing on the page seems less consistent. The input and buttons are in an `input-group`, a construct offered by Bootstrap specifically to tack buttons onto a text field without any spacing. Of course... we actually did the `input-group` wrong. It's supposed to look like this (with no inner border radius):



I didn't hear much in the way of complaints following the 2019 Roadeo. Just to be clear, you pulled the view up on your phone and were having difficulty tapping the desired button?

dirschn:
So on round-three, `ngrok` seems to work just fine. But when I attempt running it for the roadeo app, i get a blocked host error. it says to add the url to `config.hosts` but doesn't tell me where to do that (not to mention the URL changes every time I start the server over). Any pointers?

dirschn:
> The spacing on the page seems less consistent

Okay, that's fair, and I addressed that.

> The input and buttons are in an input-group, a construct offered by Bootstrap specifically to tack buttons onto a text field without any spacing

I'm sure there's something to be said for that but (and this may be a bit naive) I'm sure it's fine for use to add a little personalized flair as "needed", right? 

> you pulled the view up on your phone and were having difficulty tapping the desired button?

I have now, and when I'm very closely paying attention to what I'm pressing it's fine but I can imagine during the actual maneuver it may be beneficial for judges to be able to watch the participant and haphazardly tap the button, rather than have to look down at the screen every time they want to mark a cone.

All that said, I won't be too upset about this getting nixed, but I can see it being a slight improvement that people don't know would be helpful until they see it.

benmelz (changes requested):
I have to back @werebus on this one

> I'm sure there's something to be said for that but (and this may be a bit naive) I'm sure it's fine for use to add a little personalized flair as "needed", right?

There's a difference between personalized flair and going against the intended purpose of a component. Like he said, the core purpose of an input group _is_ to have no spacing. _If_ we want spacing, we should stop using an input group.

That being said if the problem is "these buttons are too close together", the solution I would go with would be to move the `-` button to the left side of the input.

Something like this (hacked together cause I don't have this app set up):

Screen Shot 2022-04-20 at 3 04 38 PM

werebus:
Hmm.... that either-side look looks pretty good, IMO.

dirschn:
> Like he said, the core purpose of an input group is to have no spacing. If we want spacing, we should stop using an input group.

So I just spent some time looking at the docs for `input-group`s and yeah after reading that I understand better now _what_ they are and what their purpose is. I think after reading up on that I agree that either side is definitely a better option than adding some spacing.

werebus (changes requested):
So, while we're here, can we address how we've been doing `input-group`s wrong? That is, here's a simplified version of the HTML on the page currently:

```html
<div class="input-group">
  <button class="input-group-append">...</button>
  <input>
  <button class="input-group-append">...</button>
</div>
```

And here's how it "should" be:

```html
<div class="input-group">
  <div class="input-group-prepend">
    <button>...</button>
  </div>
  <input>
  <div class="input-group-append">
    <button></button>
  </div>
</div>
```

Which will (among other things) get rid of the border radii on the sides of the buttons that touch the input.

werebus (changes requested):
Because the DOM structure changed, now disabled buttons don't un-disable when you change the value.

See [here](https://github.com/umts/umasstransit.rodeo/blob/5f17a91a6d813c955ef61bf0c740ef3b0b4ca8b4/app/assets/javascripts/scoring.js#L5), but rather than look for siblings (or now, I guess, cousins), it would be better to just look for fields that share the same `data-field`.

werebus:
Also, I think [this CSS rule](https://github.com/umts/umasstransit.rodeo/blob/5f17a91a6d813c955ef61bf0c740ef3b0b4ca8b4/app/assets/stylesheets/scoring.scss#L7) can be removed now.

dirschn:
> Because the DOM structure changed, now disabled buttons don't un-disable when you change the value.
> 
> See [here](https://github.com/umts/umasstransit.rodeo/blob/5f17a91a6d813c955ef61bf0c740ef3b0b4ca8b4/app/assets/javascripts/scoring.js#L5), but rather than look for siblings (or now, I guess, cousins), it would be better to just look for buttons that share the same `data-field`.

I am rather lost here in the search for how to fix this, any advice/guidance would be greatly appreciated

dirschn:
So I changed [this line](https://github.com/umts/umasstransit.rodeo/blob/5f17a91a6d813c955ef61bf0c740ef3b0b4ca8b4/app/assets/javascripts/scoring.js#L5) to be ` $('.scoring').find('button.increment').prop('disabled', false);` and now it does sorta work, in that clicking any of the enabled buttons will enable any disabled button. 

I'm guessing the fix for that is by choosing only buttons that change the same input, which I'm also guessing is what you meant by "buttons that share the same data-field"

dirschn:
This is the version I'm ending on for the day, it doesn't work at all but I'm hoping it's maybe close? not sure.
`$(this).find(attr('data-field="'+$(this).attr('data-field')+'"]')).prop('disabled', false);`

werebus:
Let's see if I can point you in the right direction. Here are some various statements that may help:

* [Data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) are a means of attaching data to HTML elements. In the case of the +/- buttons, each of them has one that says what field it's "for". This is a pretty common use case for data attributes: to provide some additional context to JavaScript about what type of thing just triggered an event. It's important to mention here that data attributes are just regular HTML attributes that start with `data-`
* Even though it's "just" an HTML element, both Rails views and jQuery have shortcuts for accessing them. In this JS, we're using the jQuery method; `$(something).data('field')` returns the value of `something`'s `data-field` attribute. Worth noting that we already have it stored in a variable - `field_name`. (note, jQuery *actually* does [something more complicated and confusing](https://api.jquery.com/data/#data-html5): it has it's own internal data store and initializes that store with the values from the 'data-` fields. But we can ignore that for now, just file it away as a future "gotcha")
* Both CSS and JS use the same dialect for [selecting elements](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes). There are specific syntaxes for selecting based on the "`id`" attribute (`#`) and the "`class`" attribute (`.`), but there's also one for selecting based on [any attributes at all](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) (which, of course, includes data attributes).
* In JavaScript, what `this` is can be a little overwhelming to read up on, but this code is actually making use of a jQuery-specific behavior where, in the context of a callback function, it represents the element that triggered the callback. So, within this function, `$(this)` is a jQuery object representing the button that just got `click`ed on.
* In jQuery, `find()` is part of a [collection of DOM-traversal functions](https://api.jquery.com/category/traversing/tree-traversal/). it returns chlidren of the function target. If you just want something anywhere on the page, you use the `$("some_selector") function.
* What I was recommending in my previous comment, is that we don't need to start looking for the other button starting from `this`, it will be sufficient to just select and enable everything on the whole page that has `field_name` as a `data-field` attribute.

dirschn:
> Let's see if I can point you in the right direction. Here are some various statements that may help:

Yes, this was helpful, thank you

dirschn:
Hmm..... it seems as though if I click `+`, then use the up/down arrow in the box OR manually type in a number, i can then type in 0 and `-` does NOT disable. This behavior is present in the current version of the site though. Perhaps we can just disable the input box so they _have_ to use the buttons? or just pretend we don't know it's there since it hasn't been an issue so far

app/assets/javascripts/scoring.js
```diff
@@ -2,8 +2,9 @@ $(document).ready(function(){
   $('#score_form').areYouSure();
 
   $('.scoring').on('click', 'button.increment', function(){
-    $(this).siblings('button.increment').prop('disabled', false);
     var field_name = $(this).data('field');
+    var other_button = document.querySelectorAll("[data-field='" + field_name + "']");
+    other_button.forEach(function(element){ element.disabled = false});
```
werebus, line 7: So, It's true that [you don't necessarily need to use jQuery](https://youmightnotneedjquery.com/); this is exactly how you would do this without it. And, for new projects, I am inclined _not_ to bring jQuery in because of the size and support burden.

However, in the context of a jQuery callback fuction, I think it's OK to do it the jQuery way:

```suggestion
    $("[data-field='" + field_name + "']").prop('disabled', false);
```

werebus (changes requested):
> Hmm..... it seems as though if I click +, then use the up/down arrow in the box OR manually type in a number, i can then type in 0 and - does NOT disable. I can then use the buttons to go outside the predefined range.
>
> This behavior is present in the current version of the site though. Perhaps we can just disable the input box so they have to use the buttons? Or just move on with our lives since it hasn't been an issue so far and it's not exactly a "normal" use case

I don't want to force people to use a particular input method. If you wanted to fix this problem, the "right way" would be to divorce the enabling/disabling of buttons from the changing of the input value, attaching the enable/disable code to the `change` event of the input instead.

I might do that in another PR (or even just open an issue for later), though. I'm reassured that:

* People are unlikely to mix-and-match input methods
* The mini buttons in the browser respect the min and max values
* HTML validations catches keyboard-entered values

werebus (approved):

PR #163, opened 2022-04-29
Ran rubocop

werebus:
Looks like there are a few new cops since the last time it was run.

### DIFF
--- a/app/models/circle_check_score.rb
+++ b/app/models/circle_check_score.rb
@@ -5,7 +5,7 @@ class CircleCheckScore < ApplicationRecord
 
   belongs_to :participant
 
-  validates :participant, :total_defects, :defects_found, presence: true
+  validates :total_defects, :defects_found, presence: true
   validates :participant, uniqueness: true
   validates :defects_found, numericality: {
     less_than_or_equal_to: :total_defects,
--- a/app/models/maneuver_participant.rb
+++ b/app/models/maneuver_participant.rb
@@ -20,7 +20,7 @@ class ManeuverParticipant < ApplicationRecord
   serialize :distances_achieved, Hash
 
   validates :maneuver, uniqueness: { scope: :participant }
-  validates :maneuver, :participant, :score, :reversed_direction, presence: true
+  validates :score, :reversed_direction, presence: true
 
   before_validation :set_score
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
-ruby IO.read(File.expand_path('.ruby-version', __dir__)).strip
+ruby File.read(File.expand_path('.ruby-version', __dir__)).strip
 
 gem 'bootsnap'
 gem 'bootstrap', '~> 4.3.1'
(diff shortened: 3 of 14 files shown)


app/models/maneuver_participant.rb
```diff
@@ -20,7 +20,7 @@ class ManeuverParticipant < ApplicationRecord
   serialize :distances_achieved, Hash
 
   validates :maneuver, uniqueness: { scope: :participant }
```
dirschn, line 22: Pending the answer to the above question, do we need the maneuver validation here?

app/models/circle_check_score.rb
```diff
@@ -5,7 +5,7 @@ class CircleCheckScore < ApplicationRecord
 
   belongs_to :participant
 
-  validates :participant, :total_defects, :defects_found, presence: true
+  validates :total_defects, :defects_found, presence: true
```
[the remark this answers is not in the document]
werebus: Yes, it's redundant as of Rails 5.0. `belongs_to :participant` automatically adds a validation for participant presence. If you _don't_ want to validate presence of a "belongs to", you can say `belongs_to :participant, optional: true`

app/models/maneuver_participant.rb
dirschn: So following the answer [here](https://github.com/umts/umasstransit.rodeo/pull/163#discussion_r865950883), is this validation separate from validating the presence of a maneuver? Namely, it seems to me, that it's validating this maneuver only exists once in the scope of each participant?

werebus: This is a different kind of validation; it's a uniqueness validation. Presence is _also_ validated (although the explicit validation was removed on the line below this one). There are, essentially three things that are validated with respect to the relationship between `Maneuver`s, `Participant`s, and `ManeuverParticipant`s:

1. Every `ManeuverParticipant` has a `maneuver_id` (of a `Maneuver` that exists)
2. Every `ManeuverParticipant` has a `participant_id` (of a `Participant` that exists)
3. Each combination of `maneuver_id` and `participant_id` exists only once among `ManeuverParticipant`s.

---

BTW this could also be written:

```ruby
validates :participant, uniqueness: { scope: :maneuver } 
````

The only difference is how the error message is presented — "Maneuver has already been taken" vs "Participant has already been taken".

dirschn (approved):

app/models/maneuver_participant.rb
werebus: Documentation for `validates` is [here](https://api.rubyonrails.org/v6.1/classes/ActiveModel/Validations/ClassMethods.html#method-i-validates). Two things that commonly trip newer folks up:

1. There is also a method, `validate`, (singular) which is usually not what you want.
2. There are a lot of different ways to format the arguments to `validates`. But if you write it in a way that "seems" right for what you want, it'll probably work. That flexibility is achieved via some truly arcane options munging in the Rails code, that can pretty much be hand-waved through.

I'd also recommend a skim through the [Rails Guide on validation](https://guides.rubyonrails.org/active_record_validations.html).

PR #169, opened 2022-07-19
Update Rails to 6.1.6.1 - CVE-2022-32224

werebus:


app/controllers/maneuver_participants_controller.rb
```diff
@@ -40,7 +40,7 @@ def show
 
   def update
     attrs = params.permit(:reversed_direction, :speed_achieved,
-                          :made_additional_stops, :completed_as_designed)
+                          :made_additional_stops, :completed_as_designed).to_hash
     attrs[:obstacles_hit] = parse_obstacles
     attrs[:distances_achieved] = parse_distance_targets
```
werebus, line 45: The CVE in question pertains to how serialization to/from YAML works. Even though `parse_obstacles` and `parse_distance_targets` both return real, honest-to-goodness `Hash`es — which are safe to load — putting them _into_ this `HashWithIndifferentAccess` turns them into `HashWithIndifferentAccess`es as well :grey_question: — which is _not_ on the safe list.

dirschn (approved):
seems fine to me

PR #189, opened 2023-12-12
Scoreboard pizazz

### LINKED ISSUE #102: Visual indication when scoreboard is updated
Having implemented #101, we now have some idea of _what_ data is new when the scoreboard is updated. Perhaps some sort of visual "pizazz" to let the viewer know that a score changed?

dirschn:
closes #102 

It won't animate a second time, if the same cell is updated twice in a row. I'm not entirely sure why but that's also not a very common use case at all, so I'm not super worried about it?

### DIFF
--- a/app/assets/javascripts/channels/circle_check_scores.coffee
+++ b/app/assets/javascripts/channels/circle_check_scores.coffee
@@ -4,8 +4,6 @@ App.circle_check_scores =
       participant = data.participant_id
       score = data.score
       cell = $("tr[data-participant-id=#{participant}] td.circle-check-score")
-      cell.text(score).attr('data-text', score).attr('data-score', score).addClass('last-updated')
+      cell.text(score).attr('data-text', score).attr('data-score', score)
       $("table.scoreboard").trigger("recalculate")
-      setTimeout ->
-        cell.removeClass('last-updated')
-      , 5000
+      flashCell(cell)
--- a/app/assets/javascripts/channels/circle_check_scores.coffee
+++ b/app/assets/javascripts/channels/circle_check_scores.coffee
@@ -6,3 +6,6 @@ App.circle_check_scores =
       cell = $("tr[data-participant-id=#{participant}] td.circle-check-score")
       cell.text(score).attr('data-text', score).attr('data-score', score).addClass('last-updated')
       $("table.scoreboard").trigger("recalculate")
+      setTimeout ->
+        cell.removeClass('last-updated')
+      , 5000
--- a/app/assets/javascripts/channels/circle_check_scores.coffee
+++ b/app/assets/javascripts/channels/circle_check_scores.coffee
@@ -4,5 +4,5 @@ App.circle_check_scores =
       participant = data.participant_id
       score = data.score
       cell = $("tr[data-participant-id=#{participant}] td.circle-check-score")
-      cell.text(score).attr('data-text', score).attr('data-score', score)
+      cell.text(score).attr('data-text', score).attr('data-score', score).addClass('last-updated')
       $("table.scoreboard").trigger("recalculate")
--- a/app/assets/javascripts/channels/maneuver_participants.coffee
+++ b/app/assets/javascripts/channels/maneuver_participants.coffee
@@ -12,8 +12,6 @@ App.manuever_participants =
         a.text(score)
       else
         cell.text(score)
-      cell.attr('data-text', score).attr('data-score', score).addClass('last-updated')
+      cell.attr('data-text', score).attr('data-score', score)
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/scoreboard.scss
```diff
@@ -21,3 +21,13 @@
     }
   }
 }
+
+td.last-updated {
+  animation: 1s linear 0s 5 blinkCell
```
frothedoatmilk, line 26:
```suggestion
  animation: 0.75s ease-in-out 0s 3 blinkCell
```
the `ease-in-out` is lesser to the 0.75s/3 reps change

app/assets/javascripts/channels/circle_check_scores.coffee
```diff
@@ -4,5 +4,5 @@ App.circle_check_scores =
       participant = data.participant_id
       score = data.score
       cell = $("tr[data-participant-id=#{participant}] td.circle-check-score")
-      cell.text(score).attr('data-text', score).attr('data-score', score)
+      cell.text(score).attr('data-text', score).attr('data-score', score).addClass('last-updated')
```
frothedoatmilk, line 7: To ensure it triggers on a second update, you need to remove and re-add the class so the animation will trigger again. Perhaps a method for this will help? If it's an infrequent use case then maybe too much work.

frothedoatmilk (commented):
Both minor changes, looks good otherwise!

app/assets/stylesheets/scoreboard.scss
dirschn: So the scoreboard is literally a TV that's just posted and people walk by and glance at it. I think shortening it too much would effectively counteract adding it all as no one will ever see it. 

Sure for `ease-in-out`, i don't really see a difference

app/assets/javascripts/channels/circle_check_scores.coffee
dirschn: I feel like there will be opinions about what I did... but it works!

app/assets/javascripts/channels/maneuver_participants.coffee
```diff
@@ -12,5 +12,8 @@ App.manuever_participants =
         a.text(score)
       else
         cell.text(score)
-      cell.attr('data-text', score).attr('data-score', score)
+      cell.attr('data-text', score).attr('data-score', score).addClass('last-updated')
       $("table.scoreboard").trigger("recalculate")
+      setTimeout ->
+        cell.removeClass('last-updated')
+      , 5000
```
frothedoatmilk, line 19: Abstract this update code so we can update it easily later

frothedoatmilk (changes requested):

app/assets/javascripts/channels/maneuver_participants.coffee
dirschn: Yeah good call my b

frothedoatmilk (approved):
Looks good!

dirschn:
Before merging I want to double check with @werebus and/or @benmelz that using `setTimeout` is okay... I know there are generally negative feelings towards using functions like that.

benmelz:
> Before merging I want to double check with @werebus and/or @benmelz that using `setTimeout` is okay... I know there are generally negative feelings towards using functions like that.

This is a valid/appropriate usage.

I think the negative feelings you're talking about have to do with the fact that `setTimeout`/`setInterval` are not actually precise with respect to their timings (i.e. "5 seconds" will not actually be _exactly_ 5 seconds in reality).

In this case, "remove the class in about 5 seconds" is perfectly fine and is the sort of light scripting work it (and javascript in general, ~~insert rant about node.js here and javascript frameworks here~~) was intended for.

werebus:
Yeah, there are also objections to its misuse. First, it predates `setInterval`, so _very_ old JS tutorials will have you setting timeouts inside of timeouts to achieve "every minute" type effects (in fact, you'll find such code on BusInfoBoard). It's also sometimes been used, pre-promise/async, with a timeout of 0 to create non-blocking code.

I also see some misguided sentiment in the internet along the lines of, "if it's a callback, it's bad."

dirschn:
I'm not sure if it's my home wifi not liking the vpn or what but I'm having trouble deploying, getting SSH connection timeouts. @werebus is something wrong with the app or with me?

werebus:
It's not up. I only pay for the droplet when we need it in April-May.

Right now, the domain just redirects to https://results.umasstransit.rodeo which is a [GH Pages static site](https://github.com/umts/results.umasstransit.rodeo/).

dirschn:
Well, that makes sense. Okay cool

PR #191, opened 2024-01-03
Locking scores

### LINKED ISSUE #61: Ability to "close" scoring?

frothedoatmilk:
Fix #61 through a new parameter for each user, `lock_scores`. Only takes effect if that user is an admin or a master of ceremonies, and is toggled on the scoreboard page. The scoreboard page will display text if the scores are locked, and who is locking them if you have appropriate permissions (admin or master of ceremonies)

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -3,8 +3,8 @@
 module Admin
   class UsersController < ApplicationController
     before_action :find_user, except: %i[index manage]
-    before_action(only: %i[destroy update]) { require_role :admin }
     before_action :catch_lock_scores, only: :update
+    before_action(only: %i[destroy update]) { require_role :admin }
 
     def approve
       @user.approve!
@@ -52,8 +52,18 @@ module Admin
       return if lock_scores.blank?
 
       @user.update lock_scores
-      redirect_to scoreboard_participants_path,
-                  notice: "Scores have been #{lock_scores[:lock_scores] == 'true' ? 'locked' : 'unlocked'}"
+      notice = case lock_scores
+               when 'true'
+                 'Scores have been locked.'
+               when 'false'
+                 if User.scoring_enabled?
+                   'Scores are still locked by other users.'
+                 else
+                   'Scores have been unlocked.'
+                 end
+               end
+
+      redirect_to scoreboard_participants_path, notice:
     end
   end
 end
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -4,6 +4,7 @@ module Admin
   class UsersController < ApplicationController
     before_action :find_user, except: %i[index manage]
     before_action(only: %i[destroy update]) { require_role :admin }
+    before_action :catch_lock_scores, only: :update
 
     def approve
       @user.approve!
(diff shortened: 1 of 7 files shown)


frothedoatmilk:
There aren't any view restrictions, eg you can still navigate to `http://localhost:3000/maneuver_participants` and look around, but edits are blocked unless you're an admin

frothedoatmilk:
Also if this looks good I'll make specs then, I don't really want to write them if we decide to implement differently

app/views/participants/welcome.haml
```diff
@@ -7,7 +7,6 @@
   - if Participant.any?
     To view the scoreboard, click
     = link_to 'here', scoreboard_participants_path
-    or on the button above.
```
dirschn, line 10: naughty naughty

app/controllers/admin/users_controller.rb
```diff
@@ -45,5 +46,24 @@ def user_params
...
+                 'Scores have been locked.'
+               when 'false'
+                 if User.scoring_enabled?
+                   'Scores have been unlocked.'
+                 else
+                   'Scores are still locked by other users!'
+                 end
+               end
+
+      redirect_to scoreboard_participants_path, notice:
```
dirschn, line 66: stupid ruby 3.1 I was gonna comment "silly louis you forgot something" but no, it is I who has been silly

app/helpers/scoreboard_helper.rb
```diff
@@ -43,4 +43,14 @@ def score_cell(record, new:, edit: nil, allowed: false)
...
+
+  def format_locking_users(locking_users)
+    locking_users = locking_users.pluck(:name)
+    if current_user.lock_scores?
+      locking_users.delete(current_user.name)
+      "#{locking_users.join(', ')} and you"
+    else
+      locking_users.to_sentence
+    end
+  end
```
dirschn, line 55: image
Take a look at this: https://www.rubydoc.info/docs/rails/Array:to_sentence

dirschn (changes requested):
I will say, looking at this, I think that using the `Competition` class or whatever could end up being a lot cleaner. Not necessarily against this, especially since you did all the work, but this method feels a little hacky to me. I also don't think there should be "multiple" people locking it or unlocking it, I think it should either be locked or not. There's currently no way to override the locking even if you're an admin which does not sit well with me

app/views/participants/welcome.haml
frothedoatmilk: If you're not signed in and you're on this page, there is no button! That was my only reasoning behind this, perhaps a little out of scope

app/helpers/scoreboard_helper.rb
frothedoatmilk: Yeah I was trying to not have your name be there but seems not smart

frothedoatmilk:
I spent like 10 minutes typing out a response but ultimately I didn't know that singleton was a thing so I will be making a `Competition` class

frothedoatmilk:
For now, unless the `Competition` class implodes, this PR is kinda moot

PR #192, opened 2024-01-05
Add settings to lock scores

### LINKED ISSUE #61: Ability to "close" scoring?

frothedoatmilk:
Fixes #61 using a singleton instance of the new `Settings` model, very simple controller action tied to one form on the scoreboard page.


app/controllers/settings_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class SettingsController < ApplicationController
+  def update
+    settings = Settings.instance
+    if settings.update settings_params
+      redirect_to scoreboard_participants_path,
+                  notice: settings.scores_locked? ? t('.scores_locked') : t('.scores_unlocked')
+    else
+      redirect_to scoreboard_participants_path, notice: settings.errors.full_messages.to_sentence
```
dirschn, line 10:
```suggestion
      redirect_to scoreboard_participants_path, danger: settings.errors.full_messages.to_sentence
```

app/controllers/settings_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class SettingsController < ApplicationController
+  def update
+    settings = Settings.instance
+    if settings.update settings_params
+      redirect_to scoreboard_participants_path,
+                  notice: settings.scores_locked? ? t('.scores_locked') : t('.scores_unlocked')
```
dirschn, line 8:
```suggestion
                  success: settings.scores_locked? ? t('.scores_locked') : t('.scores_unlocked')
```

config/locales/en.yml
```diff
@@ -31,3 +31,6 @@
 
 en:
   hello: "Hello world"
+  settings:
+    scores_locked: 'Scores have been locked.'
+    scores_unlocked: 'Scores have been unlocked.'
```
dirschn, line 36:
```suggestion
    scores_locked: 'Scores have been locked'
    scores_unlocked: 'Scores have been unlocked'
```
required^                loosely optional v
```suggestion
    scores_locked: 'Successfully locked scores'
    scores_unlocked: 'Successfully unlocked scores'
```

app/models/settings.rb
```diff
@@ -0,0 +1,11 @@
...
+
+require 'singleton'
+
+class Settings < ApplicationRecord
+  validates :singleton_guard, inclusion: { in: [0] }
+
+  def self.instance
+    first_or_create!(singleton_guard: 0)
+  end
+end
```
dirschn, line 11:
```suggestion
class Settings < ApplicationRecord
  include Singleton
end
```

I think this is what you want

db/migrate/20240105205316_add_settings.rb
```diff
@@ -0,0 +1,10 @@
+class AddSettings < ActiveRecord::Migration[6.1]
+  def change
+    create_table :settings do |t|
+      t.integer :singleton_guard
```
dirschn, line 4:
```suggestion
```

db/migrate/20240105205316_add_settings.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.boolean :scores_locked, default: false
+    end
+
+    add_index :settings, :singleton_guard, unique: true
+  end
```
dirschn, line 9:
```suggestion
    end
  end
```

dirschn (changes requested):

app/models/settings.rb
werebus: Have you tried it?

In particular, the `Singleton` module makes `#allocate` private which prevents instantiating by way of loading from the database. That is, you can instantiate a new `Singleton` via `.instance` and even save it to the database, but you can't start a new Ruby process and retrieve that `Singleton from the database.

dirschn: I have a working demo on my machine. I sent it to Louis, I'm not sure when he'll push. I can push up a separate branch now if you really want to see it.

dirschn: Never mind, @frothedoatmilk  found that he can still create multiple `Settings` instances and the instance doesn't persist across server restarts, like you mentioned. I hadn't tested that. Sigh.

frothedoatmilk: It also still allows as many copies of Settings as you want through `Settings#create`

frothedoatmilk:
The only thing I have left is to enforce user permissions on the action so some rogue character doesn't just curl the link and unlock the scores

app/models/settings.rb
frothedoatmilk: I love when GitHub web doesn't show me comments and I repeat something someone has already said 🥰

app/models/settings.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Settings < ApplicationRecord
+  validates :singleton_guard, inclusion: { in: [0] }
```
dirschn, line 4:
```suggestion
  validates :singleton_guard, inclusion: { in: [0] }, uniqueness: true
```

I think you'll want this as well, sort of a guard for future folks working on this in case they forget to use `instance`... Also it just matches the functionality you're going for.

app/views/participants/scoreboard.haml
```diff
@@ -1,7 +1,19 @@
 - if @participants_exist
   .scoreboard
-    %button.btn.btn-primary.fullscreen
-      %i.fas.fa-expand
+    .d-flex.align-items-center
```
dirschn, line 3: This might be splitting hairs but I think you might want to use row/cols here instead? I think this is to keep the buttons grouped on the same line?

dirschn (commented):
i accidentally clicked single comment before :(

app/views/participants/scoreboard.haml
frothedoatmilk: It is, but my understanding with row/cols is to keep items evenly spaced/things that will span a whole row, whereas here it's just two buttons

frothedoatmilk:
Thoughts on system spec? Model spec won't really do anything here

dirschn:
> Thoughts on system spec? Model spec won't really do anything here

I'm not sure, I generally end up deciding the opposite of what @benmelz and @werebus  decide on specs so I usually just defer to them

app/models/settings.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Settings < ApplicationRecord
+  validates :singleton_guard, inclusion: { in: [0] }, uniqueness: true
+
```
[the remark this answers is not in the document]
frothedoatmilk: I wonder if there's a way to delegate everything to instance...

db/schema.rb
```diff
@@ -135,6 +141,7 @@
     t.boolean "master_of_ceremonies", default: false, null: false
     t.boolean "admin", default: false, null: false
     t.boolean "approved", default: false
+    t.boolean "lock_scores", default: false
```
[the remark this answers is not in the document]
frothedoatmilk: yeah, I must've forgotten to rollback at some point

frothedoatmilk:
New thing I learned: in order to delegate class methods, you have to call it within the `class << self` block!

werebus (commented):
Test failures are of two classes:

1. System tests that record scores in some way, but the scoreboard defaults to locked. You could change that default, or unlock it in the `before(:suite)`.
2. `current_user` not being available in the `ScoreboardRenderer` (`lib/scoreboard_renderer.rb`) context.

werebus:
> `current_user` not being available in the `ScoreboardRenderer` (`lib/scoreboard_renderer.rb`) context.

See #195

frothedoatmilk:
> you could change that default

Nick explained to me that the unlocking of the scores -- in some manner -- is apparently how they begin the Roadeo. I'll just unlock in the `before(:suite)`

spec/spec_helper.rb
```diff
@@ -19,6 +19,7 @@
 
   config.before :all do
     FactoryBot.reload
+    Settings.first_or_create!(scores_locked: false)
```
werebus, line 22: Probably better to use `instance`. In particular, this won't work if someone already has a `Settings` with locked scores in their test database (like I do from previously running the specs).

```suggestion
    Settings.instance.update!(scores_locked: false)
```

werebus (changes requested):

spec/spec_helper.rb
werebus: Oh, and this should go in the `rails_helper.rb` (since it technically requires the Rails environment).

app/views/participants/scoreboard.haml
```diff
@@ -1,7 +1,19 @@
...
+      - if current_user&.role?(:master_of_ceremonies)
+        = form_with url: toggle_scores_lock_path, local: true do |f|
+          = f.button class: 'btn btn-primary mr-2' do
+            - if Settings.scores_locked?
+              Unlock Scores
+              %i.fas.fa-lock-open
+            - else
+              Lock Scores
+              %i.fas.fa-lock
+      - elsif Settings.scores_locked?
```
werebus, line 15: I'd like to skip this on the archive...

```suggestion
      - elsif Settings.scores_locked? && !@archiving
```

app/controllers/settings_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SettingsController < ApplicationController
+  def toggle_scores_lock
+    settings = Settings.instance
+    if settings.update({ scores_locked: !settings.scores_locked? })
```
werebus, line 6:
```suggestion
    if settings.toggle(:scores_locked).save
```

werebus (changes requested):

werebus (approved):

PR #196, opened 2024-01-19
Switch from pry to irb/debug

werebus:
The newest IRB has pretty good [debugger integration](https://st0012.dev/whats-new-in-ruby-3-2-irb).

As a side-effect, putting `irb` in the `Gemfile` gets us the newest console now.


dirschn (approved):
Cool

PR #198, opened 2024-01-19
Upgrade Ruby version to 3.3

werebus:
I always feel pretty daring on this app; skip right over 3.2.


dirschn (approved):
yeah, let's be daring!

PR #207, opened 2024-03-04
Change authorization warning to use a `alert` flash type

### LINKED ISSUE #205: "You are not authorized" flash uses `:notice` instead of `:alert`
Not a big thing, but here, in `application_controller.rb`:
https://github.com/umts/umasstransit.rodeo/blob/main/app/controllers/application_controller.rb#L21

It makes an access problem look chipper rather than something you should be paying attention to!

frothedoatmilk:
Fix #205, yes there's a typo and yes I did force push to fix it

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,7 +18,7 @@ class ApplicationController < ActionController::Base
   end
 
   def deny_access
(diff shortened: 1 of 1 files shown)


werebus (approved):
shur

PR #213, opened 2024-04-23
First pass at building a production docker image

werebus:
I used Fly's [dockerfile-rails](https://github.com/fly-apps/dockerfile-rails) gem to generate almost all of this PR, almost no customization yet.

It builds as long as there's a "production" database it can talk to (still TBD whether that's a problem).

Also, still TBD: how we feel about a 148K-line vendored copy of Yarn...


benmelz (commented):
While I am generally opposed to the vendoring of yarn, it'd be fine with me if this got shoved through just to see if we can get docker running. <sub>nevermind that I don't really care all that much about this app</sub>

That being said, I am a little confused...

Am I understanding that the dockerfile is... installing the same version of yarn through npm globally, then making it so that when global yarn is invoked, global yarn will... read `.yarnrc` and switch to vendored yarn?

I guess now that I've typed it all out it does seem really weird/redundant

werebus:
> Am I understanding that the dockerfile is... installing the same version of yarn through npm globally, then making it so that when global yarn is invoked, global yarn will... read `.yarnrc` and switch to vendored yarn?

Yeah, that's right. I guess the goal is to make sure that the local _dev_ version of yarn matches? The generator also [reads the string it plopped into `package.json`](https://github.com/fly-apps/dockerfile-rails/blob/5c8e349bc56149ba4dddbe8fc1b736de6d488fb4/lib/generators/dockerfile_generator.rb#L1055-L1078) in order to keep the `Dockerfile` version and the `yarn set version` version in sync.

If we aren't concerned with the gem generator continuing to work, we can depart from it more...

benmelz (approved):
I shall leave it your hands (for this app at least)

PR #214, opened 2024-04-25
Remove config/secrets

werebus:
The only "secret" we have is the SECRET_KEY_BASE. In the absence of this file, dev and test will make one for us, prod will still use `ENV["SECRET_KEY_BASE"]` if it's defined.

For now, I'm keeping the `ENV` approach in production: If I end up getting docker/kamal to work then ENV variables are way easier. If not, I _may_ switch to storing it in a credentials file.


benmelz (approved):

config/initializers/devise.rb
```diff
@@ -5,9 +5,8 @@
   config.case_insensitive_keys = [:email]
   config.strip_whitespace_keys = [:email]
   config.skip_session_storage = [:http_auth]
-  config.stretches = Rails.env.test? ? 1 : 10
+  config.stretches = Rails.env.test? ? 1 : 12
   config.pepper = '0761ef0892f647b6d3c525ee7d98043efadaae4b42d77102b56beba91bd44360d4cfb0ffa713778c281afe99e18c22ed0364330c28c0292fc25251c3f8858331'
-  config.secret_key = '7dce324450fa7e8371325c63e65afb6c14fcde6f14ab5577d16d91d0a0f590e643219d31fbbe0e314d1888a0baa3aa3393a3d7870b4ff02a9e32cbcb2febe0fa'
```
werebus, line 10: Yikes. This is AFAIK a real secret.

If you don't set it, Devise uses `secret_key_base`. Changing it now means we invalidate users' outstanding password-reset/unlock tokens. But, given we don't have a working mail provider, no one has any of those they can use.

config/initializers/devise.rb
```diff
@@ -5,9 +5,8 @@
   config.case_insensitive_keys = [:email]
   config.strip_whitespace_keys = [:email]
   config.skip_session_storage = [:http_auth]
-  config.stretches = Rails.env.test? ? 1 : 10
+  config.stretches = Rails.env.test? ? 1 : 12
   config.pepper = '0761ef0892f647b6d3c525ee7d98043efadaae4b42d77102b56beba91bd44360d4cfb0ffa713778c281afe99e18c22ed0364330c28c0292fc25251c3f8858331'
```
werebus, line 9: Having the `pepper` here in a public repo makes it useless from a security-standpoint. But it's not _worse_ than not having one (which is the default). And removing/changing it invalidates existing hashed passwords.

PR #264, opened 2025-01-06
Upgrade jquery to 3.x

werebus:


app/assets/javascripts/application.js
```diff
@@ -1,4 +1,4 @@
-//= require jquery
+//= require jquery3
```
werebus, line 1: This is an interim setup. Next is removal of "jquery_ujs", after which we won't need the `jquery-rails` Gem and can switch to getting jQuery from npm.

app/assets/javascripts/participants.js
```diff
@@ -1,12 +1,11 @@
-$(document).ready(function(){
+$(function(){
   if($('.participants .uniqueness-warning').is(':visible'))
     $('.uniqueness-warning').hide();
 
   $('.participants').on('change', 'input[type=number]', function(){
     var existingNumbers = $('.participants input[type=number]').not($(this)).map(function(){
       return parseInt($(this).val());
     });
-    existingNumbers = $.unique(existingNumbers);
```
werebus, line 9: This is deprecated, but also, an array doesn't need to be sorted or unique in order to check it for the presence of an element.

benmelz (approved):

PR #270, opened 2025-01-07
Remove coffescript

werebus:
Only the ActionCable channels were written in coffeescript, converted them to JS and removed the `coffee-rails` gem.


frothedoatmilk (approved):

frothedoatmilk (changes requested):
wait...no...something is wrong

frothedoatmilk (approved):
Sorry, had to clear my cache to get it to load. Looks good!

PR #278, opened 2025-01-08
Move number button enable/disable to a 'change' listener

### LINKED ISSUE #157: plus/minus buttons don't enable/disable when other input methods are used
If someone uses the browser built-in number field buttons or the keyboard or the mouse scroll-wheel or whatever to enter a number, the plus/minus buttons don't enable/disable as expected.

Not a huge deal, folks are unlikely to use _multiple_ input methods. But, there's probably a more correct way to do it. The enabling/disabling of those buttons should be tied to the `change` event of the field in question rather than the `click` event of the buttons themselves. Note that the current `click` handler [_triggers_ the `change` element](https://github.com/umts/umasstransit.rodeo/blob/5f17a91a6d813c955ef61bf0c740ef3b0b4ca8b4/app/assets/javascripts/scoring.js#L13) for us already.

werebus:
Closes #157


frothedoatmilk (approved):
JavaScript my beloved

PR #292, opened 2025-01-13
Small next/previous participant fixes

werebus:
1. There was no ordering on these, it just assumed that id-order and number-order were the same
2. Rubocop now says `.where(number: ...n)` is better than `.where('number < ?', n)`
3. My pet-peve: relying on variables in non-executed branches being initialized to `nil`.


frothedoatmilk (approved):
Confused as to why `where(number: n...)` doesn't generate `number > n`... but looks good!

werebus:
> Confused as to why `where(number: n...)` doesn't generate `number > n`

It's because, for example:

```ruby
range = 1...10
range.include? 10
#=> false
range.include? 1
#=> true
```

Three-dot ranges can be used to simulate "less than" because they don't contain the end value - `(1...10)` is $\left[1, 10\right)$. But they always contain the start value and therefor can't be used to stand in for "greater than", only "greater than or equal".

PR #311, opened 2025-01-27
Add GH action for Rubocop

werebus:
(now that we've got no offenses).


app/controllers/health_controller.rb
frothedoatmilk, line 1: Merge master and this should go away, then I'm good with this!

frothedoatmilk (approved):

PR #314, opened 2025-01-27
Remove never-used spinner gif

werebus:
Added in 8a9b480d049f35b0b05e78ff0d4a9412087850ec, only reference to it removed in 619dd3cc4a87f48592346cd88708130e757c2afa (two days later).


frothedoatmilk (approved):
RIP...

PR #319, opened 2025-01-31
Shore up coverage

frothedoatmilk:
Just some small changes, mainly to specs, to push coverage as high as we can go. The only things not covered right now is the error branch on `SettingsController#toggle_scores_lock` (since there are no errors possible, maybe we should error if the current user isn't MC?) and the `ApplicationJob`/`ApplicationMailer` files (since we don't have any jobs or mailers!)


werebus (approved):

werebus:
(Of course, don't forget to commit the coverage file)

PR #388, opened 2025-04-21
Upgrade ruby to 3.4.3

werebus:


benmelz (changes requested):
https://github.com/umts/umasstransit.rodeo/blob/db386419d8d8c0fd61a9de599fc04a4697e331b2/Dockerfile#L4

benmelz (approved):

PR #460, opened 2025-08-28
Drop device registration JS

### LINKED ISSUE #262: `fontcolor` is deprecated
https://github.com/umts/umasstransit.rodeo/blob/525c009ccbf158e293a138ef15ce9362384b033b/app/assets/javascripts/devise_registration.js#L4-L5

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fontcolor

This works by creating a `<font>` tag, which itself is also deprecated (with the advent of HTML 4 in ~1999)

frothedoatmilk:
Fix #262, we've been leaning away from client-side validations for a while

### DIFF
--- a/app/views/devise/registrations/edit.html.haml
+++ b/app/views/devise/registrations/edit.html.haml
@@ -14,7 +14,6 @@
   .form-group
     = f.label :password_confirmation
     = f.password_field :password_confirmation, autocomplete: 'off', class: 'form-control'
-  #check-password-match
   .form-group
     = f.label :current_password
     = f.password_field :current_password, autocomplete: 'off', class: 'form-control'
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/devise_registration.js
```diff
@@ -1,21 +0,0 @@
-function checkPasswordMatch() {
-    var password = $("#user_password").val();
-    var confirmPassword = $("#user_password_confirmation").val();
-    var noMatch = "Passwords do not match!".fontcolor('red');
-    var yesMatch = "Passwords match.".fontcolor('green');
-
-    if (password != confirmPassword){
-      $("#check-password-match").html(noMatch);
```
werebus, line 8: This empty `div` can be removed from the views, too.

werebus (changes requested):

PR #474, opened 2025-09-18
Swap to npm from yarn

frothedoatmilk:
Attempted copy of https://github.com/umts/screaming-dinosaur/pull/626


dirschn:
Didn't we just switch from yarn to npm on round three in the last year?

PR #477, opened 2025-09-23
Switch to trilogy

frothedoatmilk:
Part of the larger tooling sweep going on...


benmelz (changes requested):
I think it's appropriate to update [the current action](https://github.com/umts/umasstransit.rodeo/blob/41a9edd80c71aa47e142a9f6216b5bf8a4924437/.github/workflows/ci.yml#L22) as I don't think we've solidified if we want to move to the shared actions with or without failures in all the other toolings?

benmelz (approved):

PR #620, opened 2026-04-01
Use new shared workflow for ci

benmelz:


werebus:
Hey @benmelz, it [looks like](https://github.com/umts/umasstransit.rodeo/pull/686) this is all-green since #685 was merged.

benmelz:
Indeed, although I think I'm about to throw yet another wrench in the gears with https://github.com/umts/.github/pull/14...

I might want to smooth that over here

benmelz:
Commit-by-commit is probably a good idea due to all the oxfmt changes

oxfmt.config.ts
```diff
@@ -2,4 +2,14 @@ import { defineConfig } from "oxfmt";
...
   $schema: "./node_modules/oxfmt/configuration_schema.json",
+  ignorePatterns: [
+    "app/javascript/controllers/index.js",
+    "public/400.html",
+    "public/403.html",
+    "public/404.html",
+    "public/406-unsupported-browser.html",
+    "public/422.html",
+    "public/500.html",
+    "vendor/assets/**",
```
frothedoatmilk, line 13: It looks like the `public` files were changed in an older commit, just making sure that you still want those changes to happen? Or do you want to revert to the Rails provided files?

frothedoatmilk (changes requested):

oxfmt.config.ts
benmelz: Oopsies yes I need to revert

frothedoatmilk (approved):

PR #684, opened 2026-05-28
Store database config in credentials

werebus:


config/database.yml
```diff
@@ -0,0 +1,29 @@
+default: &default
+  url: <%= ENV.fetch('DATABASE_URL', nil) %>
```
werebus, line 2: This _appears_ to work as a general-purpose override. It does nothing when blank and I successfully did something like:

```bash
export DATABASE_URL="//rodeo:password@localhost/"
```

That is, not specifying an adapter or database name, just username, password, and host.

benmelz (approved):

